PR #4, opened 2020-06-24
Twit input

dranilew:
Currently, when the user inputs something, it does not redirect to any other parts of the website. However, for the time being, this will do for the start of the prototype.


src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Given a tweet or sentence, the servlet will extract salient terms from those
+ * and store them in a list.
+ */
+@WebServlet("/keyword")
+public class KeywordServlet extends HttpServlet {
+
+  private static final List<String> tweets = new ArrayList<>();
```
nssafir, line 20: Should we use a list data structure for now if we're starting with one tweet, or should we stick with just a single string?

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,37 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Extract saliency
+    Gson gson = new Gson();
+    String json = gson.toJson(tweets);
+    response.getWriter().println(json);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    tweets.add(request.getParameter("keyword-sentence"));
```
nssafir: Should there be some type of check that the user input is valid?

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,37 @@
...
+
```
nssafir: [Nit] delete newlines

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,37 @@
...
+    response.sendRedirect("/index.html");
+  }
+
+}
```
nssafir, line 57: [Nit] add new line

src/main/webapp/scripts/load.js
```diff
@@ -28,3 +28,18 @@ function loadObject(url) {
...
 }
+
+/**
+ * Fetches data from the keywords servlet, which includes salient terms from user input.
+ */
+function getSalientTerms() {
+  let keyTermPromise = loadObject('/keyword');
+  keyTermPromise.then((keyTerms) => {
+    console.log(keyTerms);
+
```
nssafir: [Nit] delete new line

src/main/webapp/scripts/load.js
```diff
@@ -28,3 +28,18 @@ function loadObject(url) {
...
+    // Temporary. TODO: Make some use of salient terms, such as word cloud.
+    // For now, append them to the end of the website.
+    let footer = document.getElementById('footer');
+    footer.textContent = keyTerms.join(" ");
+  });
+}
```
nssafir: [Nit] add new line

src/main/webapp/index.html
```diff
@@ -20,7 +20,12 @@
         <label for="actions-radio" id="actions-label" class="menu-label">Action</label>
       </div>
       <div id="selected-option">
-        <div id="slogan"><p>Slogan</p></div>
+        <!--<div id="slogan"><p>Slogan</p></div>-->
+        <form action="/keyword" method="POST">
+          <textarea name="keyword-sentence" style="resize: none; width: 600px; height: 350px;" placeholder="Enter your tweet or interests here"></textarea>
```
nssafir: You may want to add an external css style sheet.

src/main/webapp/scripts/load.js
dranilew: I think the new line before the comment block improves readability.

src/main/webapp/index.html
dranilew: I will put this in index.css instead.

src/main/java/com/google/sps/servlets/KeywordServlet.java
dranilew: I think for the purposes of the prototype, this would do. Perhaps I'll make it so that the textarea in which you input the tweet is pre-filled with the "hard-coded" tweet.

daperlmut:
This doesn't look like it was approved by anyone yet.  Please try to wait until there is at least one approval before requesting me/Sean to review.

src/main/webapp/index.html
```diff
@@ -20,7 +20,14 @@
...
       <div id="selected-option">
-        <div id="slogan"><p>Slogan</p></div>
+        <!--<div id="slogan"><p>Slogan</p></div>-->
+        <form action="/keyword" method="POST">
+          <textarea 
+            name="keyword-sentence" 
+            placeholder="Enter your tweet or interests here">Black Lives Matter and COVID-19.</textarea>
+          <br/>
+          <input type="submit" />
+        </form>
```
juanuribe28: I think this from should probably go in a different page. The user should log in or input the tweet in one page, and then get redirected to index.html where the keywords, news and actions are shown.

src/main/webapp/styles/index.css
```diff
@@ -57,10 +57,17 @@
...
 }
+
 #selected-option:hover {
   overflow:auto;
 }
 
+#selected-option textarea {
+  resize: none;
+  width: 600px;
+  height: 350px;
```
juanuribe28: [Optional] Use vh and vw units instead of px. This way the size of the element adapts to the size of the screen.

src/main/webapp/scripts/load.js
```diff
@@ -28,3 +28,19 @@ function loadObject(url) {
...
```
juanuribe28: I don't think this function is needed as toggle.js manages the sections' loading.

src/main/java/com/google/sps/servlets/KeywordServlet.java
nssafir: Actually just looked at the new design doc comments and I think I misunderstood, the way you have it is good.

src/main/webapp/index.html
dranilew: I will move it to the keywords page for now.

src/main/webapp/index.html
```diff
@@ -19,9 +20,7 @@
         <input type="radio" name="option" id="actions-radio" value="actions" class="hide">
         <label for="actions-radio" id="actions-label" class="menu-label">Action</label>
       </div>
-      <div id="selected-option">
-        <div id="slogan"><p>Slogan</p></div>
```
juanuribe28: Why did you remove this div?

dranilew: Oh shoot, I meant to remove the forms, not the whole thing. Mb, I'll re-add it.

juanuribe28 (approved):

juanuribe28 (approved):

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Given a tweet or sentence, the servlet will extract salient terms from those
```
daperlmut: [Nit] I think there is a minor grammar issue in this sentence which you can fix by just removing the words "from those"

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@WebServlet("/keyword")
+public class KeywordServlet extends HttpServlet {
+
+  private static final List<String> tweets = new ArrayList<>();
+  private static final List<String> terms = Arrays.asList("Black Lives Matter", "COVID-19");
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Extract saliency
```
daperlmut: [Nit] Missing a '.'

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,54 @@
...
+    response.sendRedirect("/index.html");
+  }
+
+  /**
+   * Generates and returns a JSON usable by Mustache.
+   */
+  private String mustacheJsonBuilder(Collection<String> toJson) {
+    String builder = "{\"keywords\": [";
+    for (String term : toJson) {
+      builder += "{\"term\": \"" + term + "\"},";
```
daperlmut: I think this would be much easier to read using String.format
See https://www.geeksforgeeks.org/java-string-format-examples/

Also, you can turn the builder into an actual StringBuilder object
See https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    // Remove extra comma at the end.
+    builder = builder.substring(0, builder.length() - 1) + "]}";
```
daperlmut: Just to confirm, this is Java 7 right?  I was thinking about suggestion the join function but I think it only works on lists in Java 8
https://reversecoding.net/java-8-convert-list-string-comma/

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,54 @@
...
+    return builder;
+  }
+
```
daperlmut, line 50: [Nit] Remove new line between braces

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/KeywordServlet.java
dranilew: Yea, but I feel like using the join() function makes this less intuitive/readable to the reader, since it'd look like
`String.join("},{\"term\": ", toJson);` which I feel like is harder to read than simply String.format(//stuffs here).

daperlmut: Assuming the join syntax works on lists you can just join on ",".  In Java 8 I would do this

```
List<String> terms = new ArrayList<string>();
for (String term : toJson) {
  terms.add(String.format("{\"term\": \"%s\"}", term));
}
String keywordPrefix = "{\"keywords\": [";
return String.format("{\"keywords\": [%s]}", terms.join(","))
```

You might also want to factor a function that, given a key and value return "{\"key\": value}" to make it easier to build JSON objects.  Alternatively, you can use a library that builds JSON for you.  A quick google search turns up 
this: https://docs.oracle.com/javaee/7/api/javax/json/JsonObjectBuilder.html
and this: https://github.com/stleary/JSON-java

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,53 @@
...
+    if (tweet != null && !tweet.equals("")) {
+      tweets.add(request.getParameter("keyword-sentence"));
+    }
+    response.sendRedirect("/index.html");
+  }
+
+  /**
+   * Generates and returns a JSON usable by Mustache.
+   */
+  private String mustacheJsonBuilder(Collection<String> toJson) {
```
seangilpin: I don't see this as being inherently tied to mustache. It's creating a json encoding of the terms?

src/main/webapp/json/keywords.json
```diff
@@ -1 +1,6 @@
-{"title" : "Key Terms"}
+{
+  "keyterms": [
```
seangilpin, line 3: is this meant to be keywords instead of keyterms?

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  private String jsonBuilder(Collection<String> toJson) {
+    List<String> terms = new ArrayList<>();
+    for (String term : toJson) {
+      terms.add(keyValueJson("term", term));
+    }
+
+    // Remove extra comma at the end.
+    String keywordPrefix = "{\"keywords\": [";
+    return String.format(keywordPrefix + "%s]}", String.join(",", terms));
```
daperlmut: [Nit] I include the keywordPrefix variable in my suggestion but then realized it wasn't necessary because it is only used one.  I would just hard code it here.

daperlmut: Also, I think you can use your helper function in the return here now.

`return keyValueJson("keywords", String.format("[%s]", String.join(",", terms))`

I'm still partial to using an actual JSON library for this, but you can do that in a future PR if you want.

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/KeywordServlet.java
dranilew: Yea, I realized that and removed the prefix variable, but it's not pushed yet (Might be something you see in the Saliency branch as Nathan mentioned something similar to this there.

As for the JSON library, I will probably leave that for a future PR.

daperlmut: Ok, I'll approve this PR as soon the changes to remove keywordPrefix variable are committed (I'm assuming that change is going to be made to this PR)

daperlmut (changes requested):

daperlmut (approved):

PR #7, opened 2020-06-25
Add servlet to query projects from the API. 

juanuribe28:
Add a Java Servlet to query data from the Global Giving API, and updated the html template to go accordingly.


src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private String encodeTerm(String term) throws RuntimeException {
+    try {
+        return URLEncoder.encode(term, StandardCharsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException ex) {
+        throw new RuntimeException(ex.getCause());
+    }
+  }
+
+  private String curlProjects(String apiKey, String queryTerm) throws IOException{
```
nssafir: May want to add comment here to explain what this method is doing

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    String path = "https://api.globalgiving.org/api/public/services/search/projects";
+    String queryString = String.format("?api_key=%s&q=%s", apiKey, queryTerm);
+
```
nssafir: [Nit] remove newline

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+		String[] curlCommand = { "curl", "-H", "Accept: application/json", "-H", "Content-Type: application/json", "-X", "GET", path+queryString};
```
nssafir: [Nit] unindent

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    ProcessBuilder process = new ProcessBuilder(curlCommand);
+		Process p = process.start();
```
nssafir: [Nit] unindent

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: I believe this newline increases readability by separating the formatting of the curl command from the process.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> jsonResultList = new ArrayList<>();
+    String apiKey = "API_KEY"; //Insert the API_KEY here for testing
+    for (String term : terms) {
+      String queryTerm = encodeTerm(term);
+      String jsonResult = curlProjects(apiKey, queryTerm);
+      jsonResultList.add(jsonResult);
+    }
+
```
nssafir: [Nit] delete newline

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,66 @@
...
+/** Servlet that actions items.*/
+@WebServlet("/actions")
+public class ActionsServlet extends HttpServlet {
+
+  private static final List<String> terms = Arrays.asList("Black Lives Matter", "COVID-19");
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> jsonResultList = new ArrayList<>();
+    String apiKey = "API_KEY"; //Insert the API_KEY here for testing
```
nssafir: Should this be an instance variable (asking this for myself as well)? Though I guess in the final implementation, the API key won't be in the code anyway.

juanuribe28: Yeah, I think it should be an instance variable.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+      jsonResultList.add(jsonResult);
+    }
+    String jsonResultString = "{\"results\": ["+String.join(",", jsonResultList)+"]}";
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResultString);
+  }
+
+  private String encodeTerm(String term) throws RuntimeException {
+    try {
+        return URLEncoder.encode(term, StandardCharsets.UTF_8.toString());
```
dranilew: [Nit] 2 spaces for tab

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+    } catch (UnsupportedEncodingException ex) {
+        throw new RuntimeException(ex.getCause());
```
dranilew: [Nit] 2 spaces for tab

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+      String queryTerm = encodeTerm(term);
+      String jsonResult = curlProjects(API_KEY, queryTerm);
+      jsonResultList.add(jsonResult);
+    }
+    String jsonResultString = "{\"results\": ["+String.join(",", jsonResultList)+"]}";
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResultString);
+  }
+
+  private String encodeTerm(String term) throws RuntimeException {
```
dranilew: The styling guide doesn't really say anything concerning this, but you technically don't need to add the _throws_ keyword for any unchecked exceptions, such as RuntimeException.

juanuribe28: I'll remove it.

dranilew (approved):

daperlmut:
I noticed there is no description this PR.  It is generally good practice to always include a description in your PR so it is clear what they are about.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import com.google.gson.Gson;
+
+/** Servlet that actions items.*/
+@WebServlet("/actions")
+public class ActionsServlet extends HttpServlet {
+
+  private static final List<String> terms = Arrays.asList("Black Lives Matter", "COVID-19");
+  private static final String API_KEY = "API_KEY"; //Insert the API_KEY here for testing
```
daperlmut: [Nit] Add '.' after comment

daperlmut: [Nit] 2 spaces before inline comment

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> jsonResultList = new ArrayList<>();
+    for (String term : terms) {
+      String queryTerm = encodeTerm(term);
+      String jsonResult = curlProjects(API_KEY, queryTerm);
+      jsonResultList.add(jsonResult);
+    }
+    String jsonResultString = "{\"results\": ["+String.join(",", jsonResultList)+"]}";
```
daperlmut: I mentioned this PR #4 as well but this can be read more clearly if you String.format.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+      return URLEncoder.encode(term, StandardCharsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException ex) {
+      throw new RuntimeException(ex.getCause());
+    }
+  }
+  /**
+   * Returns the a json string with the API response given a queryTerm and the API key.
+   */
+  private String curlProjects(String apiKey, String queryTerm) throws IOException {
+    String path = "https://api.globalgiving.org/api/public/services/search/projects";
```
daperlmut: [Optional] I wonder if this should be a module level constant.  I probably would do so, but I can see arguments either way so I'll leave it up to you.

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: I agree. It should be module level.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> jsonResultList = new ArrayList<>();
+    for (String term : terms) {
+      String queryTerm = encodeTerm(term);
+      String jsonResult = curlProjects(API_KEY, queryTerm);
+      jsonResultList.add(jsonResult);
+    }
+    String jsonBase = "{\"results\": [%s]}";
```
daperlmut: [Nit] Since you only use this variable once it is probably better to just hardcode.

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,66 @@
...
+      throw new RuntimeException(ex.getCause());
+    }
+  }
+  
+  /**
+   * Returns the a json string with the API response given a queryTerm and the API key.
+   */
+  private String curlProjects(String apiKey, String queryTerm) throws IOException {
+    String queryString = String.format("?api_key=%s&q=%s", apiKey, queryTerm);
+    String[] curlCommand = { "curl", "-H", "Accept: application/json", "-H", "Content-Type: application/json", "-X", "GET", API_PATH+queryString};
```
seangilpin: Doesn't java have a library for making this type of request directly?

seangilpin: something like this:
https://www.baeldung.com/java-http-request

juanuribe28: This is really useful, I'll implement it.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -0,0 +1,107 @@
...
+    List<String> jsonResultList = new ArrayList<>();
+    for (String term : terms) {
+      String queryTerm = encodeTerm(term);
+      Map<String, String> queryParameters = new HashMap<>();
+      queryParameters.put("api_key", API_KEY);
+      queryParameters.put("q", queryTerm);
+      String jsonResult = urlQuery(API_PATH, queryParameters);
+      jsonResultList.add(jsonResult);
+    }
+    String jsonResultString = String.format("{\"results\": [%s]}", String.join(",", jsonResultList));
```
daperlmut, line 42: [Nit] I mentioned this in PR #4 but there is a lot of duplicated code between different classes in terms of building JSON.  I think we should write a helper library to construct JSON, or, better yet, use an existing library such as https://docs.oracle.com/javaee/7/api/javax/json/JsonObjectBuilder.html.

This doesn't have to be done in this PR, but at the very least I would add a TODO to clean this up in the future.

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: Currently I am developing the Actions classes. Once I have them I'm planning on using GSON to convert the objects to JSON. So, I think the TODO is not necessary as this was only the prototype.

daperlmut: It's still useful to added TODOs to indicate how this will work in the future.  This gives better context to reviewers.  Maybe something like TODO(uribejaun): Factor this out into an Action class that uses GSON.

daperlmut (approved):

PR #8, opened 2020-06-26
News Servlet with Hardcoded Data

nssafir:


src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,87 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    getArticles(keywords);
+    String json = makeNewsJson();
+    System.out.println(json); // test
+    response.getWriter().println(json);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  }
```
juanuribe28: I would remove this function if it is not being used.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private String makeNewsJson() {
+    Gson gson = new Gson();
+    String json = gson.toJson(articles);
+    json = "{\"articles\": " + json + "}";  
+    return json;
+  }
+
+  private class Article {
```
juanuribe28: This class should go in com.google.sps.data.

src/main/java/com/google/sps/data/ArticleBuilder.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public ArticleBuilder withLink(String link) {
+      this.link = link;
+      return this;
+    }
+  
+    public Article build() {
+      return new Article(this.title, this.link);
+    }
+  }
```
juanuribe28: [Nit] Add newline.

src/main/java/com/google/sps/data/ArticleBuilder.java
```diff
@@ -0,0 +1,19 @@
+package com.google.sps.data;
+
+public class ArticleBuilder {
```
juanuribe28: It's better practice to have the builder inside of the parent class.

src/main/java/com/google/sps/data/Article.java
```diff
@@ -0,0 +1,11 @@
+package com.google.sps.data;
+
+public class Article {
+  private String title;
+  private String link;
+
+  public Article(String title, String link) {
```
juanuribe28: If you're going to use a builder pattern, the constructor should be private.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,49 @@
...
+    response.getWriter().println(json);
+  }
+
+  private String makeNewsJson() {
+    Gson gson = new Gson();
+    String json = gson.toJson(articles);
+    json = "{\"articles\": " + json + "}";  
+    return json;
+  }
+
```
juanuribe28: [Nit] Remove new line.

juanuribe28 (changes requested):

src/main/java/com/google/sps/data/ArticleBuilder.java
nssafir: How do you access the builder if it's inside of the parent class?

src/main/java/com/google/sps/data/ArticleBuilder.java
```diff
@@ -0,0 +1,19 @@
...
+    public String link;
+
+    public ArticleBuilder(String title) {
+      this.title = title;
+    }
+
+    public ArticleBuilder withLink(String link) {
+      this.link = link;
+      return this;
+    }
```
dranilew: It's generally also good practice to have your instance fields private (unless they're constants, in which case it mostly depends on what you're trying to do), especially since they're not _final._ 

You can also consider having an empty constructor, with another method returning ArticleBuilder that sets the title instead.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.sps.data.Article;
+import com.google.sps.data.ArticleBuilder;
+
+/**
+ * Servlet for generating news articles from keywords.
+ */
+@WebServlet("/news")
+public class NewsServlet extends HttpServlet {
+
+  private List<String> keywords = new ArrayList<String>();
```
dranilew: [Optional] You technically don't need the second <String> generic type definition here. For example:
`List<String> keywords = new ArrayList<>();`
also works, and does the same thing.

src/main/java/com/google/sps/data/ArticleBuilder.java
dranilew: As Juan mentioned in the daily sync earlier, you can basically copy-paste this whole thing into Article.java. Since it's a public class within a public class, you can simply do something like "Article.Builder builder = new Article.Builder(//params here)" to use it.

juanuribe28 (approved):

daperlmut:
Please include descriptions for your PRs

src/main/java/com/google/sps/data/Article.java
```diff
@@ -0,0 +1,29 @@
...
+    this.title = b.title;
+    this.link = b.link;
+  }
+
+  public static class Builder {
+    private String title;
+    private String link;
+
+    public Builder(String title) {
+      this.title = title;
```
daperlmut, line 23: I think you are missing a "return this;" in this function.  That will cause the method chaining to break.

daperlmut: I suppose this does depend on if you want "title" to be the only required parameter and the link parameter to be optional, in which case this design is ok.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.data.Article;
+
+/**
+ * Servlet for generating news articles from keywords.
+ */
+@WebServlet("/news")
+public class NewsServlet extends HttpServlet {
+
```
daperlmut: [Nit] Remove newline

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,46 @@
...
+  private List<String> keywords = new ArrayList<>();
+  // Hardcoded list of articles.
```
daperlmut, line 21: [Optional] I prefer to have a newline before the comments

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String json = makeNewsJson();
+    response.getWriter().println(json);
+  }
+
+  private String makeNewsJson() {
+    Gson gson = new Gson();
+    String json = gson.toJson(articles);
+    json = "{\"articles\": " + json + "}";
```
daperlmut: As noted in PRs #4 and #7, use String.format to make this more readable

daperlmut (changes requested):

src/main/java/com/google/sps/data/Article.java
nssafir: I'm a little confused about this comment -- should I put "return this;" in the builder constructor?

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,49 @@
...
+        new Article.Builder("Hearing goes off the rails when lawmaker keeps banging table",
+        "https://www.cnn.com/videos/politics/2020/06/24/louie-gohmert-bangs-table-judiciary-hearing-vpx.cnn")
+          .withDate("06/24/2020")
+          .build()
+      )   
+  );
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String json = makeNewsJson();
```
seangilpin: Ultimately the request will have keywords used to produce list of articles? If so perhaps pass articles into makeNewsJson from here.

src/main/java/com/google/sps/data/Article.java
daperlmut: Sorry, my suggestion accurate in terms of what I was think of so let me correct myself.

First of all, no, you don't need a 'return this' in the constructor.  However, in the constructor you only include parameters that are required.  Then, you want a 'with' function for each optional parameter, or for parameters that can be set at a later time.

I was considering whether you needed to include title and link as part of the constructor, or whether you wanted a separate 'withTitle' and 'withLink' function.  On second look, I think this design is fine.  You can optionally move link to it's own 'withLink' function but either way is reasonable.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+@WebServlet("/news")
+public class NewsServlet extends HttpServlet {
+  private List<String> keywords = new ArrayList<>();
+  
+  // Hardcoded list of articles. 
+  // In future implementation, will be generated based on key words.
+  private List<Article> articles = new ArrayList<>(
+      List.of(
+        new Article.Builder("Bolton says Trump turned a blind eye to the coronavirus pandemic",
```
daperlmut: [Nit] Multiline function calls should be indented by 4 spaces.  Chain function are indented by 2 spaces.  Therefore, it should look something like this (to be honest I'm not a huge fan of this formatting but it is what is recommended by the Googel Java style guide).

```
myFunction(
    parameter1,
    parameter2,
    parameter3)
  .chainedFunction(parameter)
  .chainedFunction2(parameter)
```

daperlmut (changes requested):

daperlmut (approved):

PR #9, opened 2020-06-29
Add extraction of saliency to servlet

dranilew:
This should work for pretty much all general cases. However, of course, I'll need to test this on a deployed live website as I still can't seem to get the API to work on the local dev server.


src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -17,12 +25,36 @@
 @WebServlet("/keyword")
 public class KeywordServlet extends HttpServlet {
 
-  private static final List<String> tweets = new ArrayList<>();
-  private static final List<String> terms = Arrays.asList("Black Lives Matter", "COVID-19");
+  private final List<String> tweets = new ArrayList<>();
+  private final TreeMap<Entity, Float> terms = new TreeMap<>((term1, term2) -> {
```
nssafir, line 29: [Nit] could use a comment to explain the purpose of the logic in here.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -17,12 +25,36 @@
...
+      AnalyzeEntitiesResponse entityResponse = analyzeEntity(tweet);
+      List<Entity> entities = entityResponse.getEntitiesList();
+      for (Entity entity : entities) {
+        if (entity.getSalience() == 0) {
+          continue;
+        }
+        
+        if (terms.size() < 10) {
+          terms.put(entity, entity.getSalience());
+        } else if (!terms.containsKey(entity) && entity.getSalience() > terms.get(terms.pollFirstEntry())) {
```
nssafir, line 57: Why do you need to check here if terms doesn't contain the entity but not in the if statement on line 49. If analyzeEntity only generates unique entities, you shouldn't have to check `!terms.containsKey(entity)` here. Otherwise, you would need to check `!terms.containsKey(entity)` before the previous if statement as well.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -39,10 +71,10 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
   /**
    * Generates and returns a JSON usable by Mustache.
    */
-  private String jsonBuilder(Collection<String> toJson) {
+  private String jsonBuilder(Map<Entity, Float> toJson) {
```
nssafir: [optional] Would it be better to make a new list for the json than clearing the old terms list?

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -56,5 +88,21 @@ private String jsonBuilder(Collection<String> toJson) {
   private String keyValueJson(String key, String value) {
```
nssafir: If this function is only one line, it may be better to just put this code in jsonBuilder().

src/main/java/com/google/sps/servlets/KeywordServlet.java
dranilew: This is mainly for considering multiple tweets. If the user has, say, 2 tweets that mentions "Black Lives Matter" for example, then there would be two entities with the same word. However, I'm not exactly sure how the entity hashes, as it could be the case that two different entities with the same _name_ field could hash to two completely different things.

src/main/java/com/google/sps/servlets/KeywordServlet.java
dranilew: I'm not quite understanding what you mean. Would you be able to clarify?

src/main/java/com/google/sps/servlets/KeywordServlet.java
dranilew: Ah, I guess I didn't really read the question. In the first _if_ statement, it wouldn't matter as there are less than 10 terms (for now, we're only looking at the top 10 keywords). In addition, if the hashCode() method only considers equality in _name_, then this would also simply just replace the previous salience score with the new one. However, this may or may not be desired at all, so I will add something to account for that.

nssafir: Ok, that makes sense to me.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -17,13 +25,43 @@
...
+            continue;
+          }
+          terms.put(entity, entity.getSalience());
+        } else if (!terms.containsKey(entity) && entity.getSalience() > terms.get(terms.pollFirstEntry())) {
+          terms.remove(terms.pollFirstEntry());
+          terms.put(entity, entity.getSalience());
+        } else {}
+      }
+    }
+
```
nssafir: [Nit] delete newline

nssafir (approved):

juanuribe28 (approved):

seangilpin (approved):

PR #11, opened 2020-06-30
Add projects class and implemented gson library in Actions Servlet.

juanuribe28:
Add projects class to build costumizable projects objects and reduce the loading time of the main page by reducing the json size to only essential information. At the same time, implemented the class in the Actions Servlet, along side the gson library to manipulate json.


src/main/java/com/google/sps/data/Project.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /** Class used to build projects*/
+  public static class ProjectBuilder {
+
+    private final double id;
+    private String title;
+    private String summary;
+
+    public ProjectBuilder(double id) {
```
nssafir, line 35: [optional] Should a title also be required?

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -28,19 +30,21 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<String> jsonResultList = new ArrayList<>();
+    Map<String, List<Project>> jsonResultMap = new HashMap<>();
     for (String term : terms) {
       String queryTerm = encodeTerm(term);
       Map<String, String> queryParameters = new HashMap<>();
       queryParameters.put("api_key", API_KEY);
```
nssafir: [optional] is there a more efficient way to do this so you're not putting the api key in the hashmap for every new term?

dranilew (approved):

src/main/java/com/google/sps/data/Project.java
juanuribe28: I decided to only use an id, because it is the only value that is required to query the projects from the API. So the id is all you need to get the rest of the information.

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: This is true, I'll try to update it.

src/main/java/com/google/sps/data/Project.java
```diff
@@ -0,0 +1,54 @@
+package com.google.sps.data;
+
+/** Class containing projects*/
```
seangilpin: What's a project?

seangilpin: Does it contain the project or does it represent the project?

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -28,19 +30,21 @@
...
+      List<Project> projectsList = extractProjectsList(jsonResult);
+      jsonResultMap.put(term, projectsList);
     }
-    
-    // TODO: Factor this out into an Action class that uses GSON.
-    String jsonResultString = String.format("{\"results\": {%s}}", String.join(",", jsonResultList));
+  
+    Gson gson = new Gson();
+    Map<String, Object> results = new HashMap<>();
+    results.put("results", jsonResultMap);
```
seangilpin, line 46: This is just a temporary thing?

src/main/java/com/google/sps/data/Project.java
juanuribe28: It represents a project.

juanuribe28: The API we're using returns a list of projects or initiatives that different non-profits are doing. So a project is meant to represent those.

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: What is?

seangilpin (approved):

PR #12, opened 2020-06-30
Improve frontend and update frontend to match java servlets.

juanuribe28:
Updates the templates and js to work with the new Actions and Keywords servlets (News is still hardcoded). Also, it adds a loading effect to the main screen to take into account the increased latency time resulting of using live servers instead of hardcoded values.

### DIFF
--- a/src/main/java/com/google/sps/servlets/ActionsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ActionsServlet.java
@@ -25,7 +25,7 @@ import com.google.sps.data.Project;
 public class ActionsServlet extends HttpServlet {
 
   private static final List<String> terms = Arrays.asList("Black Lives Matter", "COVID-19");
-  private static final String API_KEY = "a752ec3e-9fcf-4500-9107-694351adc5ee";  // Insert the API_KEY here for testing.
+  private static final String API_KEY = "API_KEY";  // Insert the API_KEY here for testing.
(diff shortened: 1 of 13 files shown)


nssafir (approved):

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -82,7 +82,7 @@ private String urlQuery(String basePath, Map<String, String> parameters) throws
   }
 
   /**
-   * Extracts and returns the a list of projects given a jsonStirng.
+   * Extracts and returns a list of projects given a jsonStirng.
```
daperlmut: [Nit] String is spelled wrong.

Also, a jsonString is just a string in JSON format right?  In that case I would say "given a JSON string" rather than "given a jsonString" as the latter makes it look like jsonString is some special type of class object (unless it is supposed to be a special type of object in which case I am mistaken)

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: It is a string in JSON format. I'll fix it.

daperlmut (approved):

seangilpin (approved):

PR #13, opened 2020-06-30
Implement Guardian API for News

nssafir:
Implemented the Guardian API to pull news articles from hardcoded list of keywords. There is no functional front-end component the news feature yet.

### DIFF
--- a/src/main/java/com/google/sps/servlets/NewsServlet.java
+++ b/src/main/java/com/google/sps/servlets/NewsServlet.java
@@ -21,6 +21,7 @@ import java.io.UnsupportedEncodingException;
 import java.nio.charset.StandardCharsets;
 import java.net.URLEncoder;
 import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Servlet for generating news articles from keywords.
@@ -28,7 +29,7 @@ import java.util.HashMap;
 @WebServlet("/news")
 public class NewsServlet extends HttpServlet {
 
-  private String API_KEY = "test";
+  private final String API_KEY = "test"; // Insert actual API key when testing.
   private int numArticlesPerKeyword = 3;
   private List<String> keywords = new ArrayList<String>(
     List.of("Black Lives Matter", "COVID-19")
@@ -42,6 +43,7 @@ public class NewsServlet extends HttpServlet {
       addArticlesForTerm(keyword);
     }
     String json = makeNewsJson();
+    System.out.println(json); // test
     response.getWriter().println(json);
   }
 
@@ -59,8 +61,8 @@ public class NewsServlet extends HttpServlet {
 
   private String makeNewsJson() {
     Gson gson = new Gson();
-    String json = gson.toJson(articleMap);
-    json = "{\"articles\": " + json + "}";  
+    Map<String, Object> articles = Map.of("articles", articleMap);
+    String json = gson.toJson(articles); 
     return json;
   }
 
@@ -76,17 +78,14 @@ public class NewsServlet extends HttpServlet {
     if (!status.equals("ok")) {
       System.out.println("Error: status of JSON returned from Guardian API is " + status);
       return;
-    }
-    else if (numResults < numArticlesPerKeyword) {
+    } else if (numResults < numArticlesPerKeyword) {
       if (numResults > 0) {
         resultsToShow = numResults;
-      }
-      else {
+      } else {
         System.out.println("Error: no results found for query");
         return;
       }
-    }
-    else {
+    } else {
       JsonArray storiesJsonArray = responseObject.getAsJsonArray("results");
       List<Article> articles = new ArrayList<Article>();
       for (int i = 0; i < resultsToShow; i++) {
@@ -132,12 +131,10 @@ public class NewsServlet extends HttpServlet {
       if (responseCode == 200) {
         String json = getJson(url);
         jsonToArticles(json, keyTerm);
-      }
-      else {
+      } else {
         System.out.println("Error: connection response code is: " + responseCode);
       }    
-    }
-    catch(Exception e) {
+    } catch(Exception e) {
       e.printStackTrace();
     }
   }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,135 @@
...
+      throw new RuntimeException(ex.getCause());
+    }
+  }
+
+  private String makeNewsJson() {
     Gson gson = new Gson();
-    String json = gson.toJson(articleList);
-    json = String.format("{\"articles\": %s}", json);
+    String json = gson.toJson(articleMap);
+    json = "{\"articles\": " + json + "}";
```
juanuribe28: [Optional] Instead of using string formatting, you could make it a map before passing it to gson:
```suggestion
    Map<String, Object> articles = Map.of("articles", articleMap);
    String json = gson.toJson(articles);
```

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,135 @@
...
+    JsonObject responseObject = new JsonParser().parse(jsonString).getAsJsonObject().getAsJsonObject("response");
+    // Check to see if json has valid status and enough entries.
+    String status = responseObject.get("status").getAsString();
+    int numResults = responseObject.get("total").getAsInt();
+    int resultsToShow = numArticlesPerKeyword;
+    if (!status.equals("ok")) {
+      System.out.println("Error: status of JSON returned from Guardian API is " + status);
+      return;
+    }
+    else if (numResults < numArticlesPerKeyword) {
```
dranilew: [Nit] Should be one line.
`} else if (condition) {`

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,135 @@
...
+      if (numResults > 0) {
+        resultsToShow = numResults;
+      }
+      else {
```
dranilew: Ditto above

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,135 @@
...
+        System.out.println("Error: no results found for query");
+        return;
+      }
+    }
+    else {
```
dranilew: Ditto above

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,135 @@
...
+      URL url = new URL(queryPath);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestMethod("GET");
+      connection.connect();
+      int responseCode = connection.getResponseCode();
+      if (responseCode == 200) {
+        String json = getJson(url);
+        jsonToArticles(json, keyTerm);
+      }
+      else {
```
dranilew: Ditto above

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,135 @@
...
+        System.out.println("Error: connection response code is: " + responseCode);
+      }    
+    }
+    catch(Exception e) {
```
dranilew: Ditto above

src/main/java/com/google/sps/servlets/NewsServlet.java
juanuribe28: I just realized that Map.of is only Java 9+, so you can't use it. But still I would use the Map.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,132 @@
...
-          .build(),
-        new Article.Builder(
-            "Hearing goes off the rails when lawmaker keeps banging table",
-            "https://www.cnn.com/videos/politics/2020/06/24/louie-gohmert-bangs-table-judiciary-hearing-vpx.cnn")
-          .withDate("06/24/2020")
-          .build()
-      )   
+
+  private final String API_KEY = "test"; // Insert actual API key when testing.
+  private int numArticlesPerKeyword = 3;
```
dranilew: [Nit] Since the number of articles per keyword stays constant, you could consider making this _final_ and change into a constant.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,132 @@
...
+  private List<String> keywords = new ArrayList<String>(
+    List.of("Black Lives Matter", "COVID-19")
   );
+  private HashMap<String, List<Article>> articleMap = new HashMap<String, List<Article>>();
```
dranilew: [Nit] The last `<String, List<Article>>` definition is unnecessary. You can replace it with simply `<>`

For example: `private List<String> list = new ArrayList<>();`

src/main/java/com/google/sps/servlets/NewsServlet.java
nssafir: It seemed to be working for me.

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -30,11 +30,11 @@
 public class NewsServlet extends HttpServlet {
 
   private final String API_KEY = "test"; // Insert actual API key when testing.
-  private int numArticlesPerKeyword = 3;
+  private int NUM_ARTICLES_PER_KEYWORD = 3;
```
dranilew: Since it's a constant, I recommend putting the _final_ keyword here. So:
`private final int NUM_ARTICLES_PER_KEYWORD = 3;`

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -10,42 +10,131 @@
...
+    } catch (UnsupportedEncodingException ex) {
+      throw new RuntimeException(ex.getCause());
+    }
+  }
+
+  private String makeNewsJson() {
     Gson gson = new Gson();
-    String json = gson.toJson(articleList);
-    json = String.format("{\"articles\": %s}", json);
+    Map<String, Object> articles = Map.of("articles", articleMap);
```
juanuribe28: [Nit] .of won't work on Java 8.

juanuribe28 (changes requested):

juanuribe28 (approved):

seangilpin (approved):

PR #14, opened 2020-07-01
Add Project Serlvet to serve html for the projects.

juanuribe28:
Add a java servlet to query a project given a project ID, and dynamically serve an html page containing the information. Update frontend to work with this serving.


src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.net.URLEncoder;
+import java.net.URL;
+import java.net.HttpURLConnection;
+import java.util.Map;
+
+/* Class conatining helper methods to make url requests */
+public final class UrlRequest {
```
nssafir, line 16: More a question then a comment, but why did you choose to make another java class outside of the servlet?

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,68 @@
...
+   * Returns the a json string with the API response given an URL path and the query parameters.
+   */
+  public static String urlQuery(String basePath, Map<String, String> parameters) throws IOException {
+    String path = String.format("%s?%s", basePath, getParamsString(parameters));
+    URL url = new URL(path);
+    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+    connection.setRequestMethod("GET");
+    connection.setRequestProperty("Accept", "application/json");
+    connection.setRequestProperty("Content-Type", "application/json");
+
```
nssafir: [Nit] remove whitespace

src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,39 @@
...
+  private static final String API_PATH = "https://api.globalgiving.org/api/public/projectservice/projects/%s";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Map<String, String> queryParameters = new HashMap<>();
+    queryParameters.put("api_key", API_KEY);
+    String idString = request.getParameter("id");
+    String path = String.format(API_PATH, idString);
+    String jsonString = UrlRequest.urlQuery(path, queryParameters);
+
```
nssafir: [Nit] remove whitespace

src/main/webapp/scripts/main.js
```diff
@@ -16,5 +16,11 @@ $().ready(() => {
       $(".base").show(200);
       $("#real-body").removeClass("focus");
     });
+
+    $(".active .item").click(function() {
+        $(".item").off("click");
```
nssafir: [nit] should this indentation be four spaces?

src/main/java/com/google/sps/data/UrlRequest.java
nssafir: On second look, this makes a lot of sense. I'll probably incorporate using this class into my servlets in a later PR.

src/main/webapp/scripts/main.js
juanuribe28: No, I'll fix it.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -95,45 +56,20 @@ private String urlQuery(String basePath, Map<String, String> parameters) throws
     List<Project> projectsList = new ArrayList<>();
     for (int i = 0; i < projectsListJson.size(); i++) {
       Map projectJson = (Map) projectsListJson.get(i);
```
dranilew: Raw types like used here are generally unsafe to use and should be avoided. The [Gson documentation](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/Gson.html) provides a way with which you can define parametrized types so as to avoid using raw types.

If you want to read more about raw types, there's this [StackOverflow question](https://stackoverflow.com/questions/2770321/what-is-a-raw-type-and-why-shouldnt-we-use-it) and [this tutorial](https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html) from Oracle

src/main/java/com/google/sps/servlets/ProjectServlet.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Map<String, String> queryParameters = new HashMap<>();
+    queryParameters.put("api_key", API_KEY);
+    String idString = request.getParameter("id");
+    String path = String.format(API_PATH, idString);
+    String jsonString = UrlRequest.urlQuery(path, queryParameters);
+    Gson gson = new Gson();
+    Map jsonMap = gson.fromJson(jsonString, Map.class);
```
dranilew: Same goes here as for above concerning raw types.

dranilew (approved):

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.nio.charset.StandardCharsets;
+import java.net.URLEncoder;
+import java.net.URL;
+import java.net.HttpURLConnection;
+import java.util.Map;
+
+/* Class conatining helper methods to make url requests */
+public final class UrlRequest {
+  /**
+   * Returns the a json string with the API response given an URL path and the query parameters.
```
daperlmut: [Nit] JSON should be capitalized since it is an abbreviation.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public static String urlQuery(String basePath, Map<String, String> parameters) throws IOException {
+    String path = String.format("%s?%s", basePath, getParamsString(parameters));
```
daperlmut: [Nit] I would try to be consistent with whether or not you shorten 'parameter' to 'param'.  I personally prefer to avoid shortening so I would call the function getParameterString but the shortening of 'param' is common enough that I would be ok if you use 'param' instead.  Either way, consistency is the most important thing.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,65 @@
...
+    URL url = new URL(path);
+    HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
daperlmut, line 23: [Nit] When using acronym or initialisms in CamelCase you usually upper case only the first letter.  This variable should be called 'HttpUrlConnection'.

See https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,65 @@
...
+    connection.setRequestMethod("GET");
+    connection.setRequestProperty("Accept", "application/json");
+    connection.setRequestProperty("Content-Type", "application/json");
+    int responseCode = connection.getResponseCode();
```
daperlmut: I don't think you use this anywhere right now.  What is the purpose of fetching the response code?

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,65 @@
...
+    InputStreamReader inputStream = new InputStreamReader(connection.getInputStream());
+    StringBuilder contentBuilder = new StringBuilder();
+    try (BufferedReader inputReader = new BufferedReader(inputStream)) {
```
daperlmut, line 30: If you are going to wrap this in 'try' block you should have a corresponding catch.

Also, you probably want to put connection.disconnect() in a 'finally' block to ensure you disconnect even if an exception is thrown.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,65 @@
...
+  public static String getParamsString(Map<String, String> params) throws UnsupportedEncodingException {
+      StringBuilder result = new StringBuilder();
+      for (Map.Entry<String, String> entry : params.entrySet()) {
+          result.append(URLEncoder.encode(entry.getKey(), StandardCharsets.UTF_8.toString()));
+          result.append("=");
+          result.append(URLEncoder.encode(entry.getValue(), StandardCharsets.UTF_8.toString()));
+          result.append("&");
+      }
+      String resultString = result.toString();
+      return resultString.length() > 0 ? resultString.substring(0, resultString.length() - 1) : resultString;
```
daperlmut: [Nit] Rather than doing this I would recommend the following.

Get rid of the StringBuilder and replace it with an ArrayList<String>.  Fill it with parameter values using String.format
`paramterList.add(String.format("%s=%s", ...))`

Then join on "&" at the end
`return parameterList.join("&")`

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -49,91 +44,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-          result.append("&");
-      }
-
-      String resultString = result.toString();
-      return resultString.length() > 0 ? resultString.substring(0, resultString.length() - 1) : resultString;
+  private Project getProjectFromMap(Map projectMap) {
+    double id = (double) projectMap.get("id");
+    String title = (String) projectMap.get("title");
+    String summary = (String) projectMap.get("summary");
+    Project project = new Project.ProjectBuilder(id)
```
daperlmut, line 72: When chaining methods the next line is supposed to be indented by 4 spaces for the beginning of the previous line
```
CoolObject myObject = new CoolObject.CoolObjectBuilder(requiredParam)
    .withOptionalParam(optionalParam)
```

daperlmut (changes requested):

src/main/java/com/google/sps/data/UrlRequest.java
juanuribe28: HttpUrlConnection is a class, not a variable.

src/main/java/com/google/sps/data/UrlRequest.java
juanuribe28: I was planning on adding some exception handler in case the response code is not 200. I'll add a TODO comment.

src/main/java/com/google/sps/data/UrlRequest.java
juanuribe28: Do I need the catch block even if it is a try-with-resources statement?

src/main/java/com/google/sps/data/UrlRequest.java
daperlmut: Ah, I see, this is something you imported.  Guess, there is nothing we can do about it then.

src/main/java/com/google/sps/data/UrlRequest.java
daperlmut: It depends on how you want to handle the case of the BufferedReader failing to open.  If it fails, there will be an IOException thrown.  You probably want to catch that at some point and handle it.  

I would recommend handling the Exception as soon as possible, although I'm not sure what the best design would be.  The best design depends on how this shows up in the UI and I would have to see an example of the website in action to decide how I would handle this.  For now, you might want to just create some kind of alert or pop-up to inform the user.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,74 @@
...
+  * in the Software without restriction, including without limitation the rights
+  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  * copies of the Software, and to permit persons to whom the Software is
+  * furnished to do so, subject to the following conditions:
+  * 
+  * The above copyright notice and this permission notice shall be included in all
+  * copies or substantial portions of the Software.
+  */
+
+  public static String getParamsString(Map<String, String> params) throws UnsupportedEncodingException {
```
daperlmut: This is similar to what I talked about above.  In general you want to avoid throwing errors if you can.  Think about if there is a way you can handle the Exception with having to throw an error.  Maybe you can just drop the parameter, or maybe you can drop and alert the user.  It also ok to return an HTTP error but make sure to return the correct error code.

daperlmut (changes requested):

src/main/java/com/google/sps/data/UrlRequest.java
juanuribe28: For now, I'll return an empty string.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
+
+/* Class conatining helper methods to make url requests */
+public final class UrlRequest {
+  /**
+   * Returns the a JSON string with the API response given an URL path and the query parameters.
+   */
+  public static String urlQuery(String basePath, Map<String, String> params) throws IOException {
```
daperlmut, line 20: [Nit] This no longer throws an exception right?  In that case you can remove the throws clause at the end of the function.

src/main/java/com/google/sps/data/UrlRequest.java
```diff
@@ -0,0 +1,78 @@
...
+    connection.setRequestProperty("Content-Type", "application/json");
+    int responseCode = connection.getResponseCode(); // TODO: Handle exceptions when reponse code is not 200.
+    InputStreamReader inputStream = new InputStreamReader(connection.getInputStream());
+    StringBuilder contentBuilder = new StringBuilder();
+    try (BufferedReader inputReader = new BufferedReader(inputStream)) {
+      String inputLine;
+      while ((inputLine = inputReader.readLine()) != null) {
+        contentBuilder.append(inputLine);
+      }
+    } catch(Exception e) {
```
daperlmut: [Nit] I think it is better to catch the specific type of Exception you expect.  Therefore, you should change this to `catch(IOException e)` instead of just `catch(Exception e)`

daperlmut (changes requested):

src/main/java/com/google/sps/data/UrlRequest.java
juanuribe28: The url handling and connection can still throw an IOException.

seangilpin (approved):
I think with Dan being out for a few days, it's fair to commit if you feel you properly addressed his change requests.

juanuribe28:
> I think with Dan being out for a few days, it's fair to commit if you feel you properly addressed his change requests.

It is not allowing me because Dan requested changes.

daperlmut (approved):

PR #16, opened 2020-07-01
Blob salience

dranilew:
There is no integration with the website yet, but that will come in a later PR.


src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -0,0 +1,206 @@
...
+@WebServlet("/user-image")
+public class ImageServlet extends HttpServlet {
+  
+  /**
+   * Writes an upload URL for file uploads to the servlet
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    BlobstoreService blobstore = BlobstoreServiceFactory.getBlobstoreService();
+    String uploadUrl = blobstore.createUploadUrl("/user-image");
```
juanuribe28, line 42: Just out of curiosity why are you using the server path for the upload url?

src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -0,0 +1,206 @@
...
+    ImagesService imagesService = ImagesServiceFactory.getImagesService();
+    ServingUrlOptions options = ServingUrlOptions.Builder.withBlobKey(blobKey);
+
+    // To support running in Google Cloud Shell with AppEngine's dev server, we must use the relative
+    // path to the image, rather than the path returned by imagesService which contains a host.
+    try {
+      URL url = new URL(imagesService.getServingUrl(options));
+      return url.getPath();
+    } catch (MalformedURLException e) {
+      return imagesService.getServingUrl(options);
```
juanuribe28, line 203: I think .getServingUrl doesn't work when deployed. Have you tested it?

src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Generates a list of labels that apply to the image
+   * represented by the binary data stored in imgBytes.
+   */
+  private List<EntityAnnotation> getImageLabels(byte[] imgBytes) throws IOException {
+    ByteString byteString = ByteString.copyFrom(imgBytes);
+    Image image = Image.newBuilder().setContent(byteString).build();
+
+    Feature feature = Feature.newBuilder().setType(Feature.Type.LABEL_DETECTION).build();
```
juanuribe28, line 133: Are we doing Label detection or image to text?

src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -0,0 +1,206 @@
...
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    PrintWriter writer = response.getWriter();
+    BlobKey blobKey = getBlobKey(request, "image");
+    if (blobKey == null) {
+      writer.println("Please upload an image file.");
+      return;
+    }
+    // TODO: Do something with the URL -- either add it to the writer output to the servlet, 
+    // and make use of it on the website, or remove and ignore.
+    String imageUrl = getUploadedFileUrl(request, "user-image");
```
juanuribe28, line 57: You should probably check that the file is indeed an image, or at least add a TODO or open an Issue.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -33,26 +70,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
   /**
-   * Given a key-value pair, this function returns "{key: value}".
-   */
-  private String keyValueJson(String key, String value) {
-    return String.format("{\"%s\": \"%s\"}", key, value);
+   * @return a response to the analyze-entity request
+   */ 
+  private AnalyzeEntitiesResponse analyzeEntity(String message) throws IOException {
+    try (LanguageServiceClient language = LanguageServiceClient.create()) {
```
juanuribe28: [Nit] There should be catch with the try statement.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -33,26 +70,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      Document doc = Document.newBuilder().setContent(message).setType(Document.Type.PLAIN_TEXT).build();
+      AnalyzeEntitiesRequest request = AnalyzeEntitiesRequest.newBuilder()
+          .setDocument(doc)
+          .setEncodingType(EncodingType.UTF16)
+          .build();
+      AnalyzeEntitiesResponse response = language.analyzeEntities(request);
+      language.close();
```
juanuribe28: [Nit] When doing try-with-resources you don't need to close the resource.

src/main/java/com/google/sps/servlets/ImageServlet.java
dranilew: There's a statement earlier in the method that checks whether the file is an image.

src/main/java/com/google/sps/servlets/ImageServlet.java
dranilew: Thinking about it more carefully, I guess it boils down to what kind of image the user uploads. If the user uploads a wall of text as a photo, then perhaps a text_detection is more appropriate. Wordless images would more likely need label_detection (though these labels may not be anywhere as precise as people may like them). For now, label_detection seems to be the safer move

src/main/java/com/google/sps/servlets/ImageServlet.java
dranilew: I just chose it out of whim. Does this create any side-effects or unintended functions that I should be aware of?

src/main/java/com/google/sps/servlets/ImageServlet.java
juanuribe28: I think you're checking if there's an image field, but the user could upload a non-image file to the image field.

src/main/java/com/google/sps/servlets/ImageServlet.java
juanuribe28: I'm not sure, that's why I was asking.

dranilew: Hmm I'll do some testing later when I integrate this into the website

PR #17, opened 2020-07-01
Implement Location Servlet

nssafir:
Implement a location servlet that uses Google's Geocoder API to get important features (ex. street name, city, ZIP code) from a user's location coordinates. This will later fed into the Google Civic API to get users in touch with their local politicians. Later, I will need to implement a doPost() method with Google's geolocation API to get the user's coordinates if they consent.


src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,97 @@
...
+    * Converts json response from Geolocation API to HashMap with important geographic characteristics.
+    */
+    private void jsonToMap(String jsonString) {
+      JsonObject responseObject = new JsonParser().parse(jsonString).getAsJsonObject();
+      try {
+        JsonArray resultsArray = responseObject.getAsJsonArray("results");
+        JsonArray addressArray = resultsArray.get(0).getAsJsonObject().getAsJsonArray("address_components");
+        JsonObject entry;
+        String name, geoType;
+        for (int i = 0; i < addressArray.size(); i++){
```
juanuribe28, line 77: [Optional] A comment here to explain the logic would be useful.

src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,97 @@
...
+          entry = addressArray.get(i).getAsJsonObject();
+          name = entry.getAsJsonPrimitive("long_name").getAsString();
+          geoType = entry.getAsJsonArray("types").get(0).getAsString();
+          if (geoType.equals("locality")) {
```
juanuribe28: I know switch statements are not ideal, but this seems like a a good place to use one.

src/test/java/com/google/sps/LocationTest.java
```diff
@@ -0,0 +1,25 @@
+package com.google.sps;
```
juanuribe28: I would leave this file out of this PR, as it doesn't contain any tests.

src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,104 @@
...
+        JsonArray resultsArray = responseObject.getAsJsonArray("results");
+        JsonArray addressArray = resultsArray.get(0).getAsJsonObject().getAsJsonArray("address_components");
+        JsonObject entry;
+        String name, geoType;
+        // Extract useful location informatin from JSON array.
+        for (int i = 0; i < addressArray.size(); i++){
+          entry = addressArray.get(i).getAsJsonObject();
+          name = entry.getAsJsonPrimitive("long_name").getAsString();
+          geoType = entry.getAsJsonArray("types").get(0).getAsString();
+          switch (geoType) {
```
juanuribe28, line 81: [Nit] It requires a default label, even if it is empty.

juanuribe28 (changes requested):

juanuribe28 (approved):

seangilpin (approved):

PR #21, opened 2020-07-07
Update keywords->Actions and News pipeline.

juanuribe28:
Update data pipeline to use keywords values instead of hardcoded server values. Aslo, save keywords on cookies to reduce loading time.


src/main/webapp/scripts/cookies.js
```diff
@@ -0,0 +1,85 @@
...
+  setCookie(name, value, 1);
+}
+
+
+/**
+ * Get the value of a cookie.
+ */
+function getCookie(name) {
+  let decodedCookie = decodeURIComponent(document.cookie);
+  let cookies = decodedCookie.split(';');
```
nssafir: Do you need a new variable for cookies or could you just do this operation on decodedCookie again

src/main/webapp/scripts/toggle.js
```diff
@@ -1,34 +1,59 @@
...
-  for (let i = 0; i < result.keywords.length; i++) {
-    let term = result.keywords[i].term;
-    result.keywords[i].news = objs[1].articles[term];
-    result.keywords[i].actions = objs[2].results[term];
+function renderTemplateObj(template, objs, keywords) {
+  let result = new Object();
+  result.keywords = new Array(keywords.length);
+  for (let i = 0; i < keywords.length; i++) {
+    let term = keywords[i];
+    result.keywords[i] = new Object();
```
nssafir: Can you save result.keywords[i] with 'let' like you do for term so you don't have to index on each of these lines?

src/main/webapp/scripts/cookies.js
juanuribe28: decodedCookie is a string with all the cookies and cookies is the array of strings containing all the individual cookies.

nssafir (approved):

src/main/webapp/scripts/cookies.js
```diff
@@ -0,0 +1,85 @@
+const HOURS_PER_DAY = 24;
+const MINUTES_PER_HOUR = 60;
+const SECONDS_PER_MINUTE = 60;
+const SECONDS_PER_DAY = HOURS_PER_DAY*MINUTES_PER_HOUR*SECONDS_PER_MINUTE;
```
daperlmut: [Nit] Put spaces on each side of a binary operator
e.g. `x = 2 + 2` instead of `x=2+2`

src/main/webapp/scripts/cookies.js
```diff
@@ -0,0 +1,85 @@
...
+  let name = event.currentTarget.id;
+  let value = event.currentTarget.checked;
+  setCookie(name, value, 1);
+}
+
+
+/**
+ * Get the value of a cookie.
+ */
+function getCookie(name) {
```
daperlmut: [Optional] I am used to seeing the methods defined in the order getX, setX, deleteX/eraseX since this is how most IDEs will autofill them.  There is no requirement that you move this method to the top but it makes a bit more sense in my head so I thought I'd add a suggestion in case you want to do that.

daperlmut (changes requested):

src/main/webapp/scripts/cookies.js
juanuribe28: I like that, I'll change it.

src/main/webapp/scripts/cookies.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+  return "";
+}
+
+/**
+ * Create a new cookie based on name, value and days before expiring.
+ */
+function setCookie(name, value, expirationDays) {
+  let date = new Date();
+  date.setTime(date.getTime+(expirationDays*SECONDS_PER_DAY*MILLISECONDS_PER_SECOND));
```
daperlmut: [Nit] Same note here about spaces around the binary operators.  

Also, I would personally remove the parens around the multiplication since it is well known that multiplication happens before addition in term of order of operation.  However, you may keep them if you think aid readability.

daperlmut (changes requested):

daperlmut (approved):

PR #23, opened 2020-07-07
Implement Books API

nssafir:
Transformed previous News Servlet to use Google Books API to recommend books instead. Also, changed html template to title, authors, description, and link for each book.


src/main/webapp/scripts/cookies.js
```diff
@@ -0,0 +1,85 @@
+const HOURS_PER_DAY = 24;
```
dranilew: These kinda seem like Juan's changes on top of your own in one pull request. You might want to re-create the pull request and change the base branch.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,151 @@
...
+    if (numResults < resultsToShow) {
+      if (numResults > 0) {
+        resultsToShow = numResults;
+      } else {
+        System.err.println("Error: no results found for query");
+        return;
+      }
+    }
+    JsonArray booksJsonArray = responseObject.getAsJsonArray("items");
+    List<Book> books = new ArrayList<Book>();
```
dranilew: [Nit] You do not need to define the generic type for the constructor.
`List<Book> books = new ArrayList<>();`

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,151 @@
...
+
+ /**
+  * Formats writers into string from JSON array.
+  * @return writers in String fomat.
+  */
+  private String formatWriters(JsonArray writersArray) {
+    String returnString = "";
+    int length = writersArray.size();
+    for (int i = 0; i < length - 1; i++){
+      returnString += writersArray.get(i).getAsString() + ", " ;
```
dranilew: [Nit] Remove extra space before semicolon

dranilew (changes requested):

src/main/webapp/scripts/cookies.js
nssafir: My bad, thought I did that.

src/main/webapp/scripts/toggle.js
```diff
@@ -50,7 +50,7 @@ function renderTemplateObj(template, objs, keywords) {
     let term = keywords[i];
     let keywordsObj = new Object();
     keywordsObj.term = term;
-    keywordsObj.news = objs[0].articles[term];
+    keywordsObj.news = objs[0].books[term];
```
juanuribe28: [Nit] Change .news for .books and update the mustache template accordingly.

src/main/webapp/templates/keywords.html
```diff
@@ -7,8 +7,8 @@
         <div class="section-header">News</div>
         {{#news}}
```
juanuribe28: Same as above.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Servlet for generating books from keywords.
+ */
+@WebServlet("/books")
+public class BooksServlet extends HttpServlet {
+
+  private final int NUM_BOOKS_PER_KEYWORD = 3;
```
juanuribe28: [Optional] I think we could probably use a little more books, but it's up to you.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,151 @@
...
+  private List<String> keywords = new ArrayList<String>(
+    List.of("Black Lives Matter", "COVID-19")
```
juanuribe28: [Nit] .of is not available in java 8.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Map;
+
+/**
+ * Servlet for generating books from keywords.
+ */
+@WebServlet("/books")
+public class BooksServlet extends HttpServlet {
+
+  private final int NUM_BOOKS_PER_KEYWORD = 3;
+  private List<String> keywords = new ArrayList<String>(
```
juanuribe28: [Optional] Instead of using hardcoded keywords the pipeline is working, and you can use the key query params.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,151 @@
...
+      returnString += writersArray.get(i).getAsString() + ", ";
+    }
+    returnString += writersArray.get(length - 1).getAsString();
+    return returnString;
+  }
+
+ /**
+  * Adds a Book object in books for a key term.
+  */
+  private void addBooksForTerm(String keyTerm, String encodedKeyTerm) {
```
juanuribe28: [Optional] Instead of passing both arguments, you could pass only keyTerm and encode it within the method.

juanuribe28 (changes requested):

src/main/webapp/scripts/toggle.js
nssafir: Good call -- I also changed anything with the word 'news' in it to 'books' in the html div's and main.css file.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -20,7 +20,7 @@
 @WebServlet("/actions")
 public class ActionsServlet extends HttpServlet {
 
-  private static final String API_KEY = "API_KEY";  // Insert the API_KEY here for testing.
+  private static final String API_KEY = "a752ec3e-9fcf-4500-9107-694351adc5ee";  // Insert the API_KEY here for testing.
```
juanuribe28: [Nit] Remove key

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,148 @@
...
+public class BooksServlet extends HttpServlet {
+
+  private final int NUM_BOOKS_PER_KEYWORD = 5;
+  private String[] keywords;
+  private LinkedHashMap<String, List<Book>> booksMap = new LinkedHashMap<>();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    keywords = request.getParameterValues("key");
+    for (String keyword : keywords) {
```
juanuribe28: [Optional] Use a lambda expression to evaluate every value without a for loop.
https://stackoverflow.com/questions/7221833/how-can-i-call-a-method-on-each-element-of-a-list/20684006#20684006

juanuribe28 (changes requested):

juanuribe28 (approved):

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+    booksMap.put(keyTerm, books);
+    System.out.println(booksMap); // test
+  }
+
+ /**
+  * Reads in JSON from an open URL.
+  * @return JSON in String format.
+  */
+  private String getJson(URL url) throws IOException {
```
seangilpin: I would call it read instead of get, perhaps readJsonFile

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+  private String makeBooksJson() {
+    Gson gson = new Gson();
+    Map<String, Object> books = new HashMap<>();
+    books.put("books", booksMap);
+    String json = gson.toJson(books); 
+    return json;
+  }
+
+ /**
+  * Adds Book object to books from JSON results of Guardian query.
```
seangilpin: should this be google books?

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    String queryParam = "q=" + encodeTerm(keyTerm); 
+    String queryPath = path + queryParam;
+    try {
+      URL url = new URL(queryPath);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestMethod("GET");
+      connection.connect();
+      int responseCode = connection.getResponseCode();
+      if (responseCode == 200) {
+        String json = getJson(url);
```
seangilpin: It would be helpful if the variable name described the nature of the json content.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+      JsonObject bookJson = booksJsonArray.get(i).getAsJsonObject().getAsJsonObject("volumeInfo"); 
+      String title = bookJson.get("title").getAsString();
+      String link = bookJson.get("infoLink").getAsString();
+      String image = bookJson.getAsJsonObject("imageLinks").get("thumbnail").getAsString();
+      String description = bookJson.get("description").getAsString();
+      String writer = formatWriters(bookJson.getAsJsonArray("authors"));
+      books.add(new Book.Builder(title, link).withImage(image).withDescription(description).withWriter(writer).build());
+    }
+    booksMap.put(keyTerm, books);
+    System.out.println(booksMap); // test
```
seangilpin: Does this mean you intend to remove this line later?

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    Map<String, Object> books = new HashMap<>();
+    books.put("books", booksMap);
+    String json = gson.toJson(books); 
+    return json;
+  }
+
+ /**
+  * Adds Book object to books from JSON results of Guardian query.
+  */
+  private void jsonToBooks(String jsonString, String keyTerm) {
```
seangilpin: The name isn't informative. Your adding the <keyterm/book list> to the global index/cache. The name should reflect that more closely.

src/main/java/com/google/sps/servlets/BooksServlet.java
nssafir: My bad, should have been removed.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    JsonObject responseObject = new JsonParser().parse(jsonString).getAsJsonObject();
+    // Check to see if json has enough entries.
+    int numResults = responseObject.get("totalItems").getAsInt();
+    int resultsToShow = NUM_BOOKS_PER_KEYWORD;
```
seangilpin: i would simplify this a bit by just using something like:
numResults = min(numResults, NUM_BOOKS_PER_KEYWORD)
if numResults <= 0 {
  System.err.println("Error: no results found for query");
  return;
}

src/main/java/com/google/sps/servlets/BooksServlet.java
seangilpin: In general this comment could be more descriptive

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+  */
+  private String encodeTerm(String term) {
+    try {
+      return URLEncoder.encode(term, StandardCharsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException ex) {
+      throw new RuntimeException(ex.getCause());
+    }
+  }
+
+  private String makeBooksJson() {
```
seangilpin: encodeBookMapAsJson?

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+public class BooksServlet extends HttpServlet {
+
+  private final int NUM_BOOKS_PER_KEYWORD = 5;
+  private List<String> keywords;
+  private LinkedHashMap<String, List<Book>> booksMap = new LinkedHashMap<>();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    keywords = Arrays.asList(request.getParameterValues("key"));
+    keywords.forEach((keyword) -> addBooksForTerm(keyword));
```
seangilpin: if the term has already been looked up in a previous request you don't need to do this?

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    String json = makeBooksJson();
```
seangilpin: This appears to produce a list from terms across all requests? If thats not what you want maybe filter by the keywords from this specific request.

src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,106 @@
+package com.google.sps.servlets;
```
seangilpin: I didn't see anything about the locationservlet in the description for this CL? Any reason why it being added here?

seangilpin (changes requested):

src/main/java/com/google/sps/servlets/BooksServlet.java
nssafir: I'm a little confused about this comment and the comment below. Should the keywords class variables not be refreshed not be reset every time doGet is called?

src/main/java/com/google/sps/servlets/LocationServlet.java
nssafir: That's odd, not sure why it's here.

src/main/java/com/google/sps/servlets/BooksServlet.java
seangilpin: Yeah keywords is replaced but booksMap is not is my understanding.

nssafir: Oh I see, I should make sure to clear booksMap.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Map;
+
+/**
+ * Servlet for generating books from keywords.
+ */
+@WebServlet("/books")
+public class BooksServlet extends HttpServlet {
+
+  private final int NUM_BOOKS_PER_KEYWORD = 5;
+  private List<String> keywords;
```
seangilpin: You have this as a class variable but its not clear to me why as its only ever used in the same function that initializes it.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Map;
+
+/**
+ * Servlet for generating books from keywords.
+ */
+@WebServlet("/books")
+public class BooksServlet extends HttpServlet {
+
+  private final int NUM_BOOKS_PER_KEYWORD = 5;
+  private LinkedHashMap<String, List<Book>> booksMap;
```
seangilpin: So now this doesn't need to be a class member either. The intent and flow of code will be more clear if you instantiate it directly in doGet and pass into the functions that need to modify or read it.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+import com.google.gson.JsonElement;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.net.URLEncoder;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Servlet for generating books from keywords.
```
seangilpin: I wouldn't understand this comment if I didn't already know what you are trying to do. Don't be afraid to have long function descriptions. If its a trivial function or one that wouldn't benefit from describing in detail, then short descriptions like this are more appropriate.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+  private String encodeBookMapAsJson() {
+    Gson gson = new Gson();
+    Map<String, Object> books = new HashMap<>();
+    books.put("books", booksMap);
+    String json = gson.toJson(books); 
+    return json;
+  }
+
+ /**
+  * Adds Book object to books from JSON results of Google Books query.
```
seangilpin: - It's adding more than one book.
- What is "books" (the thing you are adding Book object to)?
- Your ultimately modifying this bookmaps thing which isn't mentioned.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+  */
+  private void addBooksToMap(String jsonString, String keyTerm) {
+    JsonObject responseObject = new JsonParser().parse(jsonString).getAsJsonObject();
+    // Check to see if json has enough entries.
```
seangilpin: can you just use:
numResults = booksJsonArray.size()

it would make it easier for me to make the relationship between these two variables.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+    int numResults = responseObject.get("totalItems").getAsInt();
+    numResults = Math.min(numResults, NUM_BOOKS_PER_KEYWORD);
+    if (numResults <= 0) {
```
seangilpin: Do you need to do this. I assume there will be many terms that produce no results so you need to gracefully deal with that. If you just removed this whole if block the function would put <'term, [ empty ] > into bookMap and then whatever code has to present that information to the user can figure out how to tell the user that term had no results.

What do you think?

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+    jsonReader.close();
+    return data;
+  }
+
+ /**
+  * Formats writers into string from JSON array.
+  * @return writers in String fomat.
+  */
+  private String formatWriters(JsonArray writersArray) {
```
seangilpin: I would call this authors. I got confused and thought it meant writers in the technical sense like writing a file.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+    String returnString = "";
```
seangilpin: Can you use String.join(",", list) to do the same thing

src/main/java/com/google/sps/servlets/BooksServlet.java
seangilpin: I still feel like this comment is too terse and not informative enough.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+    returnString += writersArray.get(length - 1).getAsString();
+    return returnString;
+  }
+
+ /**
+  * Adds books objects for a key term.
+  */
+  private void addBooksForTerm(String keyTerm) {
+    String path = "https://www.googleapis.com/books/v1/volumes?";
```
seangilpin, line 123: factor out the part that does the querying of the api into another function

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+    String queryPath = path + queryParam;
+    try {
+      URL url = new URL(queryPath);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestMethod("GET");
+      connection.connect();
+      int responseCode = connection.getResponseCode();
+      if (responseCode == 200) {
+        String jsonString = readJsonFile(url);
+        addBooksToMap(jsonString, keyTerm);
```
seangilpin: This one line is very different from all of the rest of the code in this function and is hard to find. You can factor it out by having this function just return the json and then passing that json into addBooksToMap in the parent function

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,142 @@
...
+      returnString += writersArray.get(i).getAsString() + ", ";
+    }
+    returnString += writersArray.get(length - 1).getAsString();
+    return returnString;
+  }
+
+ /**
+  * Adds books objects for a key term.
+  */
+  private void addBooksForTerm(String keyTerm) {
```
seangilpin: am i confused or is the same thing being inconsistently referred to as: term, keyterm, keyword through all of the code

src/main/java/com/google/sps/servlets/BooksServlet.java
nssafir: String.join(",", list) doesn't work if list is a JsonArray, and there doesn't seem to be a very elegant way to convert a JsonArray to a Java List.

src/main/java/com/google/sps/servlets/BooksServlet.java
nssafir: I think you're right -- the Mustache template handles it well when there are no results, so there's not really a reason to print an error if there are no results.

src/main/java/com/google/sps/servlets/BooksServlet.java
seangilpin: Did you try this:
Gson googleJson = new Gson();
ArrayList jsonObjList = googleJson.fromJson(jsonArr, ArrayList.class);

source: https://stackoverflow.com/questions/18544133/parsing-json-array-into-java-util-list-with-gson

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,148 @@
...
+    try {
+      URL url = new URL(queryPath);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestMethod("GET");
+      connection.connect();
+      int responseCode = connection.getResponseCode();
+      if (responseCode == 200) {
+        return readJsonFile(url);
+      } else {
+        System.err.println("Error: connection response code is: " + responseCode);
```
seangilpin, line 138: if (responseCode != 200) { 
  System.err.println("Error: connection response code is: " + responseCode);
}
return readJsonFile(url);

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,148 @@
...
+    int length = authorsArray.size();
+    for (int i = 0; i < length - 1; i++){
+      returnString += authorsArray.get(i).getAsString() + ", ";
+    }
+    returnString += authorsArray.get(length - 1).getAsString();
+    return returnString;
+  }
+
+ /**
+  * Gets string of JSON of Google Books API's response to a query including the term.
```
seangilpin: It also makes the request to google books api.

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,148 @@
...
+import com.google.gson.JsonElement;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.net.URLEncoder;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Servlet for generating an array of book objects for each term.
```
seangilpin: a map of each term to a list of the corresponding related books

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -0,0 +1,148 @@
...
+ * The get function writes a mapping of terms to lists of book objects in the response.
```
seangilpin: I guess maybe this should be the javadoc comment for the get function

src/main/java/com/google/sps/servlets/BooksServlet.java
nssafir: Thanks, I didn't see Gson had a method for that. Would it be appropriate to declare a Gson object as a class variable if I use it in two different methods?

seangilpin (approved):
Thanks for making all those changes :)

dranilew (approved):

PR #26, opened 2020-07-09
Implement Civic Servlet in frontend.

juanuribe28:
Add the necessary javascript to call the civic servlet whenever the results page loads. Currently it is hardcoded to always make the call, and to use hardcoded coordinates. Also, it is missing the html and css for the section.


src/main/webapp/scripts/toggle.js
```diff
@@ -40,25 +43,55 @@ async function loadHtmlSections(templateUrl, objsUrls, keywordsPromise) {
...
+    result.location = locationObj;
   }
   let htmlSections = Mustache.render(template, result);
   return htmlSections;
 }
 
+/**
+ * Extract the government officials from the civic API response.
+ */
+function extractOfficials(civicObj) {
```
nssafir: Is there any reason you're doing this in javascript instead of making a more tailored JSON in the civic servlet? This method seems different than the other methods in toggle.js.

juanuribe28: I decided to do it in js because it was easier to manipulate due to its structure. However, I do agree that it seems different than the other methods in toggle.js, so I'll add it to a different file.

src/main/webapp/scripts/civic.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Extract the government officials from the civic API response.
+ */
+function extractOfficials(civicObj) {
+  let officials = new Object();
+  let levels = ["administrativeArea1", "administrativeArea2", "country", "international", "locality", "regional", "special", "subLocality1", "subLocality2"];
```
nssafir: Should this be a const?

src/main/webapp/scripts/civic.js
```diff
@@ -0,0 +1,19 @@
...
+  levels.forEach((level) => {
+    officials[level] = new Array();
+  });
+  civicObj.offices.forEach((officeObj) => {
```
nssafir, line 20: Could use some commenting in the block below this, it was a bit hard for me to figure out what was going on.

src/main/webapp/scripts/civic.js
juanuribe28: Yeah

src/main/webapp/scripts/civic.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Extract the government officials from the civic API response.
+ */
+function extractOfficials(civicObj) {
+  let officials = new Object();
+  const levels = ["administrativeArea1", "administrativeArea2", "country", "international", "locality", "regional", "special", "subLocality1", "subLocality2"];
```
dranilew: [Nit] Use CONSTANT_CASE for constant variables.

src/main/webapp/scripts/toggle.js
```diff
@@ -1,9 +1,12 @@
 const KEYWORDS_TEMPLATE_URL = '/templates/keywords.html';
 
 const KEYWORDS_OBJ_URL = '/json/keywords.json';
+
+const CIVIC_OBJ_URL = '/actions/civic'
+
```
dranilew: [Nit] Remove vertical whitespace.

src/main/webapp/scripts/civic.js
```diff
@@ -0,0 +1,20 @@
...
+  levels.forEach((level) => {
+    officials[level] = new Array();
+  });
+  civicObj.offices.forEach((officeObj) => {
+    const level = officeObj.levels[0];
```
dranilew, line 21: [Nit] CONSTANT_CASE

dranilew (changes requested):

src/main/webapp/scripts/civic.js
juanuribe28: This isn't a module-local (top-level) constant.

src/main/webapp/scripts/civic.js
juanuribe28: This isn't a module-local (top-level) constant.

dranilew (approved):

src/main/webapp/scripts/civic.js
daperlmut: [Nit] This is longer than 80 characters.  I would put each entry on its own line.

`
const level = [
  "adminArea1"
  "adminArea2"
  "country"
   ...
]`

daperlmut: Huh, the code format didn't respect my newlines.  

const level = [ 
  "adminArea1,
  "adminArea2",
   "country",
  ... 
]

daperlmut: And the regular format doesn't respect leading newlines...

Oh well, pretend each entry in the list is intended by 2 spaces.

daperlmut (changes requested):

daperlmut (approved):

PR #27, opened 2020-07-09
Implement location cookies

nssafir:
Use Navigator API to create location cookies if user chooses to share location and implement front-end to ask user. The cookies do not affect the request to the civic servlet yet (that will implemented in the next PR).


src/main/webapp/index.html
```diff
@@ -16,6 +17,9 @@
       <div id="form-title">
         <p>Copy your Tweet here</p>
       </div>
+      <label><input type="checkbox" onclick="setLocationCookie()" /> Use Location</label>
```
juanuribe28: [Nit] Input tag should be outside the label.

src/main/webapp/index.html
```diff
@@ -6,6 +6,7 @@
     <link rel="stylesheet" href="/styles/base.css">
     <link rel="stylesheet" href="/styles/index.css">
     <link href="https://fonts.googleapis.com/css2?family=Open+Sans&family=Oswald&display=swap" rel="stylesheet">
+    <script src="scripts/cookies.js"></script>
```
juanuribe28: [Optional] Move the script tag to the end of the body, as it isn't needed during loading.

src/main/webapp/index.html
juanuribe28: [Optional] Is best practice to keep javascript separate from markup, so I would remove the onclick attribute and instead add a event listener to javascript, or even better use jQuery.

src/main/webapp/scripts/cookies.js
```diff
@@ -82,3 +82,19 @@ function loadCheckboxStateCookie(cookieName) {
...
+ */
+function setLocationCookie() {
+  if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition(function(position) {
+        setCookie("latitude", position.coords.latitude, 1);
+        setCookie("longitude", position.coords.longitude, 1);
+      }, function(err) {
+        console.warn(`ERROR(${err.code}): ${err.message}`);
+        document.getElementById("location-warning")
+            .innerHTML = "We cannot access your location. Try checking your browswer settings";
```
juanuribe28: [Optional] Instead of changing the DOM I would send an alert to the user.

juanuribe28 (changes requested):

src/main/webapp/index.html
nssafir: Which javascript file would this be in?

juanuribe28: index.js. If it doesn't exist, create it and add the function.

nssafir: This is the only way I am able to get the label and checkbox to appear inline with each other on the website. Let me know if you know another way.

juanuribe28: You can use the for attribute https://www.w3schools.com/tags/tag_label.asp

nssafir: That's still not working

juanuribe28: That's weird

src/main/webapp/styles/index.js
```diff
@@ -0,0 +1 @@
+$("checkbox").click(setLocationCookie());
```
juanuribe28: [Nit] remove callback function parentheses and add new line.

juanuribe28 (changes requested):

juanuribe28 (approved):

daperlmut (approved):

PR #30, opened 2020-07-10
Add UI for civic information.

### LINKED ISSUE #38: Government Officials is not center on portrait orientation.
When switching from landscape to portrait orientation the section's tittle comes down.

juanuribe28:
Add HTML and CSS for the civic information provided by the civic servlet. The profiles of the government officials still use a hardcoded description and are missing their specific UI design.


nssafir (approved):

src/main/webapp/scripts/civic.js
```diff
@@ -2,19 +2,31 @@
...
+  for (const level in levels) {
+    let levelsObj = new Object();
+    levelsObj.levelType = level;
+    levelsObj.levelName = levels[level]
+    levelsObj.officials = new Array();
+    levelsArray.push(levelsObj);
+  }
   civicObj.offices.forEach((officeObj) => {
     const level = officeObj.levels[0];
     // Loop through the officials Indices that work at the current office, and add them to their coresponding level.
```
daperlmut: [Nit] I think this is clear from the code since you are using a forEach loop and the variable is name officialIdex.  In generally try to avoid comments that say _what_ the code does.  Comments should instead be used to explain _why_ the code is written in the way it is.

daperlmut (changes requested):

daperlmut (approved):

src/main/webapp/scripts/civic.js
```diff
@@ -2,19 +2,30 @@
...
+                  "subLocality1" : "Sublocality 1",
+                  "special" : "Special",};
+  for (const level in levels) {
+    let levelsObj = new Object();
+    levelsObj.levelType = level;
+    levelsObj.levelName = levels[level]
+    levelsObj.officials = new Array();
+    levelsArray.push(levelsObj);
+  }
   civicObj.offices.forEach((officeObj) => {
```
seangilpin, line 26: can these two foreach's be accomplished with a regular for loop like you have above, or is there a specific reason why you're doing it this way here¿

src/main/webapp/scripts/civic.js
```diff
@@ -2,19 +2,30 @@
  * Extract the government officials from the civic API response.
  */
```
seangilpin, line 7: Describe the structure of the returned levelsArray

src/main/webapp/scripts/toggle.js
```diff
@@ -65,7 +65,7 @@ async function renderTemplateObj(template, objs, keywords) {
     let civicObj = await loadObject(`${CIVIC_OBJ_URL}?lat=${lat}&lng=${lng}`);
     let locationObj = new Object();
     locationObj.address = civicObj.normalizedInput;
-    locationObj.officials = extractOfficials(civicObj);
+    locationObj.levels = extractOfficials(civicObj);
```
seangilpin: maybe something like officialsByLevel

src/main/webapp/scripts/civic.js
juanuribe28: I prefer using forEach because I believe is more readable, but I had to use a regular for loop above because I was iterating over an object and not an array.

PR #31, opened 2020-07-10
Add datastore integration with keywords

dranilew:
Please note that all the keyword datastore functionality resides in the Keyword class, which is in the data package. Refer to the class when implementing datastore keyword functionalities into your own servlets.


src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+    HashSet<String> keywordSet = new HashSet<>();
+    int index = 0;
+    for (com.google.appengine.api.datastore.Entity entity : results.asIterable()) {
+      keywordSet.add((String) entity.getProperty("keyword"));
+
+      index++;
+      if (index >= MAX_NUM_KEYWORDS) {
+        break;
+      }
+    }
```
juanuribe28: When making a query to datastore you can limit the amount of results returned, instead of doing it manually.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Adds the labels from the analysis of an image into the datastore.
+   */
+  public static void addKeywords(List<EntityAnnotation> blobAnalysis) throws IOException {
+    String labels = "";
+    for (EntityAnnotation label : blobAnalysis) {
+      labels += label.getDescription() + ", ";
+    }
```
juanuribe28: I think you could use .join(", ") here.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+  
+  /**
+   * Adds salient keywords from the message into the datastore.
+   */
+  private static void addToDatastore(String message) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    AnalyzeEntitiesResponse entityResponse = analyzeEntity(message);
+    List<Entity> entities = entityResponse.getEntitiesList();
+    System.out.println(entities.size());
```
juanuribe28: Were you printing this for testing?

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -20,11 +20,12 @@
 @WebServlet("/actions")
 public class ActionsServlet extends HttpServlet {
 
-  private static final String API_KEY = "API_KEY";  // Insert the API_KEY here for testing.
+  private static final String API_KEY = "a752ec3e-9fcf-4500-9107-694351adc5ee";  // Insert the API_KEY here for testing.
```
juanuribe28: [Nit] Remove api key

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+      datastoreEntity.setProperty("salience", entity.getSalience());
+      datastore.put(datastoreEntity);
+    }
+  }
+
+  /**
+   * @return a response to the analyze-entity request
+   */ 
+  private static AnalyzeEntitiesResponse analyzeEntity(String message) throws IOException {
+    try (LanguageServiceClient language = LanguageServiceClient.create()) {
```
juanuribe28, line 121: Add a catch statement.

src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -59,19 +57,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    }
+    Keywords.addKeywords(annotations);
+
+    Set<String> keywords = Keywords.getKeywords();
     Gson gson = new Gson();
-    String json = gson.toJson(imageLabels);
-    writer.println(json);
-    response.sendRedirect("/main.html");
+    String json = gson.toJson(keywords);
+    response.sendRedirect(String.format("/results?k=%s", Keywords.KEY));
```
juanuribe28: I'm not sure if I'm understanding correctly but keywords.KEY is a constant value, and what should be passed to the URL is the datastore key of the object, so it can be retrieved later. This is just passing a constant URL for every set of keywords.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -20,11 +20,12 @@
...
   private static final String API_PATH = "https://api.globalgiving.org/api/public/services/search/projects";
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Use key to grab keywords from datastore, or use Keywords class.
```
juanuribe28: The key value in here refers to the keywords themselves (refer to the javascript to see what is being passed to this servlet). Actions and News (Books) servlets shouldn't be accessing datstore, all the keywords are already being passed as query parameters. This way we can save the keywords in the client side, and just get the update information without having the latency of accessing datastore.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -26,79 +22,21 @@
...
-        } else {}
-      }
-    }
-    
-    // TODO: Retrieve tweets from datastore.
-    List<String> keywords = new ArrayList<>();
-    for (Entity entity : terms.keySet()) {
-      keywords.add(entity.getName());
-    }
+    Set<String> keywords = Keywords.getKeywords();
```
juanuribe28: You should be probably be passing the datstore key to this method, if not the same set of keywords would be returned to any client.

src/main/java/com/google/sps/servlets/KeywordServlet.java
```diff
@@ -26,79 +22,21 @@
...
-          .setEncodingType(EncodingType.UTF16)
-          .build();
-      AnalyzeEntitiesResponse response = language.analyzeEntities(request);
-      return response;
-    } catch (Exception ex) {
-      ex.printStackTrace();
-      return null;
+      Keywords.addKeywords(tweet);
     }
+    response.sendRedirect(String.format("/results?k=%s", Keywords.KEY));
```
juanuribe28: Same as above. Keywords.KEY is a constant and it should be case specific.

src/main/webapp/scripts/toggle.js
```diff
@@ -1,8 +1,8 @@
 const KEYWORDS_TEMPLATE_URL = '/templates/keywords.html';
 
-const KEYWORDS_OBJ_URL = '/json/keywords.json';
+const KEYWORDS_OBJ_URL = '/keyword';
 const NEWS_OBJ_URL =  '/json/news.json';
-const ACTIONS_OBJ_URL = '/actions';
+const ACTIONS_OBJ_URL = '/action';
```
juanuribe28: The servlet path is /actions in this branch.

juanuribe28 (changes requested):
Most of the comments I added are related to how the keywords pipeline is being handled here, and how the keywords are being saved in datastore. With the current changes to the previously designed pipeline there's now way to access the user specific comments.

src/main/java/com/google/sps/servlets/ImageServlet.java
juanuribe28: By object I mean the set of keywords extracted from every single input.

src/main/java/com/google/sps/servlets/ActionsServlet.java
juanuribe28: Sending the keywords to the servlets is already being handled by the fontend.

src/main/java/com/google/sps/data/Keywords.java
dranilew: Alright, I will look into this.

src/main/java/com/google/sps/data/Keywords.java
dranilew: I did consider this, but you can't really join a list of objects into a string if you need to combine one of the fields of those objects. At least, from what I know.

src/main/java/com/google/sps/data/Keywords.java
dranilew: Yes, I guess this I skimmed over this while removing all the test prints that used to be all around the class.

src/main/java/com/google/sps/servlets/ImageServlet.java
dranilew: How do you suggest creating keys for each of these sets of keywords? Because the key has to come from somewhere. As it is right now, I'm storing each of the keywords and their salience as entities via the same key (as defined by KEY). The only way I can think of right now to change this is to use user authentication and use the user ID to differentiate between each user's set of keywords.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+  public static Set<String> getKeywords() {
+    Query query = new Query(KEY).addSort("salience", SortDirection.DESCENDING);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+    
+    HashSet<String> keywordSet = new HashSet<>();
+    int index = 0;
+    for (com.google.appengine.api.datastore.Entity entity : results.asIterable()) {
+      keywordSet.add((String) entity.getProperty("keyword"));
+
```
nssafir: [Nit] delete newline

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+      labels += label.getDescription() + ", ";
+    }
+    addToDatastore(labels);
+  }
+
+  /**
+   * Adds the salient keywords from the analysis of a textual user input into the datastore.
+   */
+  public static void addKeywords(String message) throws IOException {
+    addToDatastore(message);
```
nssafir: What's the point of using this function instead of calling addToDatastore?

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+  
+  /**
+   * Adds salient keywords from the message into the datastore.
+   */
+  private static void addToDatastore(String message) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
```
nssafir: [Nit] delete newline

src/main/java/com/google/sps/servlets/ImageServlet.java
juanuribe28: Every element that is added to datastore has a unique key assigned to them.

src/main/java/com/google/sps/data/Keywords.java
dranilew: It streamlines this method with the other method that takes input from the ImageServlet to create a set of salient keywords. So instead of having to call addToDatastore() for text-inputs, and addKeywords() for the ImageServlet, it makes more sense to have both call the same method (but different parameters) to achieve the same purpose.

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -25,6 +25,7 @@
...
```
juanuribe28: [Nit] Remove comment.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,129 @@
...
+ * Provide methods for the storage, extraction, and retrieval of keywords.
+ */
+public final class Keywords {
+
+  private static final int MAX_NUM_KEYWORDS = 10;
+
+  /**
+   * @return a list of the 10 most salient keywords
+   */
+  public static List<String> getKeywords(String key) throws EntityNotFoundException {
```
juanuribe28: [Optional] Sometimes is better to handle the exception as soon as it comes up instead of throwing it. For example you could just return an empty array instead of the exception.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,129 @@
...
+    List<String> keywordList = new ArrayList<>();
+    int index = 0;
+    for (String keyword : keywordCollection) {
+      keywordList.add(keyword);
+    
+      index++;
+      if (index >= MAX_NUM_KEYWORDS) {
+        break;
+      }
+    }
```
juanuribe28: [Nit] Use a filter in the query instead of doing it manually.

src/main/java/com/google/sps/data/Keywords.java
juanuribe28: Oh yeah, that's right

juanuribe28: Here's another option: https://stackoverflow.com/questions/44245790/join-a-list-of-objects-properties-into-a-string/44245853

juanuribe28 (changes requested):

src/main/java/com/google/sps/data/Keywords.java
dranilew: I'm not using a query anymore. Instead, I'm using the entity key to directly retrieve the entity from the datastore, rather than querying it. The entity list contains the full keyword list (rather than truncating it down to 10 keywords), so I think that this is still required.

juanuribe28: Makes sense.

PR #37, opened 2020-07-10
Keyword datastore

dranilew:
This should be the working version of the datastore. THe nits will come soon.


src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+      labels += label.getDescription() + ", ";
+    }
+    return addToDatastore(labels);
+  }
+
+  /**
+   * Adds the salient keywords from the analysis of a textual user input into the datastore.
+   * @return the key pointing to the newly added datastore entity.
+   */
+  public static String addKeywords(String message) throws IOException {
```
nssafir, line 74: I think I said this in an earlier PR, but do you need to make an extra function that's just calling another function.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+    return addToDatastore(message);
+  }
+  
+  /**
+   * Adds salient keywords from the message into the datastore.
+   */
+  private static String addToDatastore(String message) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    AnalyzeEntitiesResponse entityResponse = analyzeEntity(message);
+    List<com.google.cloud.language.v1.Entity> entities = entityResponse.getEntitiesList();
```
nssafir, line 84: Maybe try to import this so you don't need to use the fully qualified name.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+    // We must create a new list of Strings as a collection of NLP Entities is not supported by datastore.
+    List<String> keywordList = new ArrayList<>();
+    for (com.google.cloud.language.v1.Entity entity : orderSet) {
+      keywordList.add(entity.getName());
+    }
+    Entity datastoreEntity = new Entity("Keyword");
+    datastoreEntity.setProperty("keywords", keywordList);
+    datastore.put(datastoreEntity);
+    return KeyFactory.keyToString(datastoreEntity.getKey());
+  }
```
nssafir, line 110: The block of code from lines 104-111 seem like what you did in the get method, could you make this another method?

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+ * Provide methods for the storage, extraction, and retrieval of keywords.
+ */
+public final class Keywords {
+
+  private static final int MAX_NUM_KEYWORDS = 10;
+
+  /**
+   * @return a list of the 10 most salient keywords
+   */
+  public static List<String> getKeywords(String key) throws EntityNotFoundException {
```
juanuribe28: Is it still throwing this exception?

src/main/java/com/google/sps/servlets/ActionsServlet.java
```diff
@@ -25,6 +25,7 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // TODO: Use key to grab keywords from datastore, or use Keywords class.
```
juanuribe28: Remove comment.

src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -52,26 +50,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     PrintWriter writer = response.getWriter();
     BlobKey blobKey = getBlobKey(request, "image");
-    if (blobKey == null) {
+    if (blobKey == null) {  // Not an image file.
```
juanuribe28: This will be null if nothing was uploaded, but still a non image file could be passed.

src/main/java/com/google/sps/data/Keywords.java
dranilew: Yea, it helps streamline the function calls so I'm not calling addToDatastore for one thing and addKeywords for another.

P3BBL35: Yea, it helps streamline the function calls so I'm not calling addToDatastore for one thing and addKeywords for another.

src/main/java/com/google/sps/data/Keywords.java
dranilew: There are two different Entity classes being used, and this is the only way to help Java differentiate between the two.

src/main/java/com/google/sps/data/Keywords.java
dranilew: They're actually quite different. This method adds all keywords into a list to put into data store, whereas the getKeywords method retrieves and puts a cap of 10 keywords into a new list for returning since if there are more than 10 keywords, we don't want more than that.

src/main/java/com/google/sps/data/Keywords.java
dranilew: Nope, I seem to have neglected to remove this.

src/main/java/com/google/sps/servlets/ImageServlet.java
dranilew: I will add a TODO for now.

juanuribe28 (approved):

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+ * Provide methods for the storage, extraction, and retrieval of keywords.
+ */
+public final class Keywords {
+
+  private static final int MAX_NUM_KEYWORDS = 10;
+
+  /**
+   * @return a list of the 10 most salient keywords
+   */
+  public static List<String> getKeywords(String key) {
```
daperlmut: [Optional] I might call the variable `keyString` instead of key to make clear that it is the string representing the key and delineate it from the key object itself.  On the other hand, this is also mostly clear from context since you can see that the variable is of type String so leaving the name as is would also be ok.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+    try {
+      Key datastoreKey = KeyFactory.stringToKey(key);
```
daperlmut: [Nit] This is only used once so I would just inline the variable.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+      List<String> keywordList = new ArrayList<>();
+      int index = 0;
+      for (String keyword : keywordCollection) {
+        keywordList.add(keyword);
+    
+        index++;
+        if (index >= MAX_NUM_KEYWORDS) {
+          break;
+        }
+      }
```
daperlmut: I believe this can be done more succinctly with the limit() method

List<String> keywordList = keywordCollection.stream().limit(MAX_NUM_KEYWORDS).collect(Collectors.toList());

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * Adds the labels from the analysis of an image into the datastore.
+   * @return the key pointing to the newly added datastore entity.
+   */
+  public static String addKeywords(List<EntityAnnotation> blobAnalysis) throws IOException {
+    String labels = "";
+    for (EntityAnnotation label : blobAnalysis) {
+      labels += label.getDescription() + ", ";
```
daperlmut: The way this is currently written will mean that labels will have a trailing ", ".  Is that intended?  If not, I would suggest making a list of label.getDescription() for each label and then calling String.join() on that list.

src/main/java/com/google/sps/data/Keywords.java
daperlmut: I can see the idea behind this but if that is the case I would recommend adding a comment explaining the reason for this function overload.  Otherwise, I think anyone who looks at this code would wonder why you are adding an extraneous layer of indirection.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,132 @@
...
+      keywordList.add(entity.getName());
+    }
+    Entity datastoreEntity = new Entity("Keyword");
+    datastoreEntity.setProperty("keywords", keywordList);
+    datastore.put(datastoreEntity);
+    return KeyFactory.keyToString(datastoreEntity.getKey());
+  }
+
+  /**
+   * @return a response to the analyze-entity request
```
daperlmut: [Nit] Needs a '.' at the end.

src/main/java/com/google/sps/servlets/ImageServlet.java
```diff
@@ -52,26 +50,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     PrintWriter writer = response.getWriter();
     BlobKey blobKey = getBlobKey(request, "image");
-    if (blobKey == null) {
+    if (blobKey == null) { // TODO: Check whether file is an image.
```
daperlmut: [Nit] Need 2 spaces after the bracket

daperlmut (changes requested):

src/main/java/com/google/sps/data/Keywords.java
dranilew: The trailing ", " does not affect the extraction of salience from the labels string, so although not really intended, it also doesn't affect the functionality of anything else in the class.

dranilew:
The most recent commit addresses all the requested changes. The max number of keywords was reduced to 5 because having 10 seems to have problems connected with the size of the JSONs returned by the other servlets and their APIs.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -0,0 +1,128 @@
...
+    for (EntityAnnotation label : blobAnalysis) {
+      labels += label.getDescription() + ", ";
+    }
+    return addToDatastore(labels);
+  }
+
+  /**
+   * Adds the salient keywords from the analysis of a textual user input into the datastore.
+   * This method overloads addKeywords() mainly to streamline how keywords are extracted
+   * and added into the keyword. Instead of having to call addToDatastore(message) for one
```
daperlmut: "how keywords are extracted and added into the keyword"
Should this say "how keywords are extracted and added into the datastore"?

src/main/java/com/google/sps/data/Keywords.java
daperlmut: Hmm, I guess that's true but I think it is still better to use String.join here since what you really _want_ is a comma separated string.  You can just create a List<String> inside the for loop and then join it at the end.

src/main/java/com/google/sps/data/Keywords.java
daperlmut: Thinking more about this I wonder if there is a better design.  It doesn't have to be done in this PR but it might be a good idea to use a more object orient style here.

Typically, when you need to overload methods to handle different classes, you could do things better by creating class methods.

e.g. Suppose I have the following code

```
public class Dog {
  private int age;
}

public class Cat {
  private float height;
}

public class Runner {
  private static void takeCareOfAnimals(List<Dog> dogs, List<Cat> cats) {
    for (Dog dog: dogs) {
      walk(dog)
    }
    for (Cat cat: cat) {
      walk(cat)
    }
  }

  private static void walk(Dog dog) {
    TimeUnit.MINUTES.sleep(5);
    system.out.println("Woof Woof");  
    TimeUnit.MINUTES.sleep(5);
  }

  private static void walk(Dog dog) {
    TimeUnit.MINUTES.sleep(5);
    system.out.println("Meow");  
    TimeUnit.MINUTES.sleep(5);
  }
}
```
The whole purpose of subclassing is to allow related objects to share properties without having to overload methods.  A better way to write the code would be as follows
```
public abstract class Animal {
  private abstract void MakeNoise();

  private void Walk() {
    TimeUnit.MINUTES.sleep(5);
    this.MakeNoise();  
    TimeUnit.MINUTES.sleep(5);      
  }
}

public class Dog extends Animal {
  private int age;

  private void MakeNoise() {
    system.out.println("Woof Woof"); 
  }
}

public class Cat {
  private float height;

  private void MakeNoise() {
    system.out.println("Meow"); 
  }
}

public class Runner {
  private static void takeCareOfAnimals(List<? extends Animal> animals) {
    for (Animal animal: animals) {
      walk(animal)
    }
  }
}
```
If the classes were a bit less related I might use an Interface (e.g. Walkable) instead of an AbstractClass but the same  principle holds.

I think you might be able to use this style to better design the code here.

daperlmut (changes requested):

daperlmut (approved):

PR #47, opened 2020-07-14
Implement Testing Framework

nssafir:
Implement framework for unit testing using JUnit and Mockito in test directory. Not comprehensive testing yet. Collection of utility functions in TestingUtil.java and example tests in BooksTest.java for Books Servlet.

### DIFF
--- a/src/test/java/com/google/sps/BooksTest.java
+++ b/src/test/java/com/google/sps/BooksTest.java
@@ -32,9 +32,7 @@ public final class BooksTest {
     booksServlet.doGet(booksGet.getRequest(), booksGet.getResponse());
     String output = booksGet.getWriter().toString();
     output = TestingUtil.getInnerMap(output);
-    System.out.println("noOutputTest() actual output is " + output);
-    System.out.println("noOutputTest() expected output is " + EMPTY_MAP_STRING);
-    Assert.assertTrue(output.equals(EMPTY_MAP_STRING));
+    Assert.assertEquals(output, EMPTY_MAP_STRING);
   }
 
  /**
@@ -50,7 +48,6 @@ public final class BooksTest {
     booksServlet.doGet(booksGet.getRequest(), booksGet.getResponse());
     String output = booksGet.getWriter().toString();
     output = TestingUtil.getInnerMap(output);
-    System.out.println("someOutputTest() output is " + output);
     Assert.assertNotEquals(output, EMPTY_MAP_STRING);
   }
   
--- a/src/test/java/com/google/sps/BooksTest.java
+++ b/src/test/java/com/google/sps/BooksTest.java
@@ -1,0 +1,57 @@
+package com.google.sps.servlets;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.net.URL;
+import org.mockito.Mockito;
+import java.io.IOException;
+import java.util.Map;
+import java.util.HashMap;
+
+/** */
+@RunWith(JUnit4.class)
+public final class BooksTest {
+  
+  public String EMPTY_MAP_STRING = "{}";
+  
+ /**
+  * Checks that if the keyword was nonsense, no result is returned.
+  */
+  @Test
+  public void noOutputTest() throws IOException {
+    Map<String, String[]> badParams = new HashMap<>();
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/sps/BooksTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.HashMap;
+
+/** */
+@RunWith(JUnit4.class)
+public final class BooksTest {
+  
+  public String EMPTY_MAP_STRING = "{}";
+  
+ /**
+  * Checks that if the keyword was nonsense, no result is returned.
```
juanuribe28, line 23: [Optional] I don't know if nonsense would be the correct word. There might be normal keywords that just don't return any results.

src/test/java/com/google/sps/BooksTest.java
```diff
@@ -0,0 +1,54 @@
...
+  */
+  @Test
+  public void noOutputTest() throws IOException {
+    Map<String, String[]> badParams = new HashMap<>();
+    String[] gibberish = {"kjdaflkdsalklkjfdsalkflkds"};
+    badParams.put("key", gibberish);
+    SetUpObject booksGet = TestingUtil.mockGet(badParams);
```
juanuribe28, line 30: I think you never import TestingUtil

src/test/java/com/google/sps/BooksTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+ /**
+  * Checks that reasonable keywords produce non-empty result.
+  */ 
+  @Test
+  public void someOutputTest() throws IOException {
+    Map<String, String[]> goodParams = new HashMap<>();
+    String[] keyWords = {"water", "air"};
+    goodParams.put("key", keyWords);
+    SetUpObject booksGet = TestingUtil.mockGet(goodParams);
```
juanuribe28, line 46: I think you don't import this class either.

src/test/java/com/google/sps/BooksTest.java
```diff
@@ -0,0 +1,54 @@
...
+    BooksServlet booksServlet = new BooksServlet();
```
juanuribe28, line 47: Or this class.

src/test/java/com/google/sps/SetUpObject.java
```diff
@@ -0,0 +1,31 @@
+package com.google.sps.servlets;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.StringWriter;
+import java.util.*;
+import org.mockito.Mockito;
+
+public class SetUpObject {
```
juanuribe28, line 9: The name should be more specific about the object type. This could be any object.

src/test/java/com/google/sps/BooksTest.java
```diff
@@ -0,0 +1,54 @@
...
+ /**
+  * Checks that if the keyword was nonsense, no result is returned.
+  */
+  @Test
+  public void noOutputTest() throws IOException {
+    Map<String, String[]> badParams = new HashMap<>();
+    String[] gibberish = {"kjdaflkdsalklkjfdsalkflkds"};
+    badParams.put("key", gibberish);
+    SetUpObject booksGet = TestingUtil.mockGet(badParams);
+    BooksServlet booksServlet = new BooksServlet();
```
juanuribe28, line 31: Does this mean that you're calling the real API during testing?

juanuribe28 (changes requested):

src/test/java/com/google/sps/BooksTest.java
nssafir: Do I need to import it if it's in the same package -- TestingUtil is another class I made with some static methods.

src/test/java/com/google/sps/BooksTest.java
nssafir: Yes. Because I made the tests in an older branch, I didn't want to test specific methods in the servlet, but for future tests it would be ideal to also test the individual methods.

src/test/java/com/google/sps/SetUpObject.java
```diff
@@ -0,0 +1,31 @@
+package com.google.sps.servlets;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.StringWriter;
+import java.util.*;
```
juanuribe28, line 6: [Nit] Do not use wildcard imports.

src/test/java/com/google/sps/BooksTest.java
juanuribe28: That's right, nevermind.

juanuribe28 (changes requested):

PR #50, opened 2020-07-15
Fix duplicate keywords

### LINKED ISSUE #49: Repeated Keywords
In some cases the same keyword appears multiple times.

dranilew:
Using a set instead of a list shoudl fix this issue. However, it still would not be able to discern between, for example, "every day" vs. "everyday".

### DIFF
--- a/src/main/java/com/google/sps/data/Keywords.java
+++ b/src/main/java/com/google/sps/data/Keywords.java
@@ -99,12 +99,12 @@ public final class Keywords {
       orderSet.add(entity);
     }
     // We must create a new list of Strings as a collection of NLP Entities is not supported by datastore.
-    Set<String> keywordList = new HashSet<>();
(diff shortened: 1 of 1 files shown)


juanuribe28 (approved):

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -99,9 +99,9 @@ private static String addToDatastore(String message) throws IOException {
       orderSet.add(entity);
     }
     // We must create a new list of Strings as a collection of NLP Entities is not supported by datastore.
-    List<String> keywordList = new ArrayList<>();
+    Set<String> keywordList = new HashSet<>();
```
juanuribe28: I would change the name to keywordSet

juanuribe28 (approved):

daperlmut (approved):

PR #54, opened 2020-07-15
Implement Testing Framework

nssafir:
Implement framework for unit testing using JUnit and Mockito in test directory. Not comprehensive testing yet. Collection of utility functions in TestingUtil.java and example tests in BooksTest.java for Books Servlet. 

This was originally PR #47, which Juan had made changes on, but for some reason the PR stopped reflecting the latest commits made to the remote branch, so I made a new a new PR.

### DIFF
--- a/src/test/java/com/google/sps/BooksTest.java
+++ b/src/test/java/com/google/sps/BooksTest.java
@@ -20,14 +20,14 @@ public final class BooksTest {
   public String EMPTY_MAP_STRING = "{}";
   
  /**
-  * Checks that if the keyword was nonsense, no result is returned.
+  * Checks that if the keyword is gibberish, no result is returned.
   */
   @Test
   public void noOutputTest() throws IOException {
     Map<String, String[]> badParams = new HashMap<>();
     String[] gibberish = {"kjdaflkdsalklkjfdsalkflkds"};
     badParams.put("key", gibberish);
-    SetUpObject booksGet = TestingUtil.mockGet(badParams);
+    ServletSetUpObject booksGet = TestingUtil.mockGet(badParams);
     BooksServlet booksServlet = new BooksServlet();
     booksServlet.doGet(booksGet.getRequest(), booksGet.getResponse());
     String output = booksGet.getWriter().toString();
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/sps/TestingUtil.java
```diff
@@ -0,0 +1,36 @@
+package com.google.sps.servlets;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.StringWriter;
+import java.util.*;
```
juanuribe28: [Nit] Don't use wildcard imports.

juanuribe28 (changes requested):

juanuribe28 (approved):

src/test/java/com/google/sps/BooksTest.java
```diff
@@ -0,0 +1,54 @@
...
+  public void noOutputTest() throws IOException {
+    Map<String, String[]> badParams = new HashMap<>();
+    String[] gibberish = {"kjdaflkdsalklkjfdsalkflkds"};
+    badParams.put("key", gibberish);
+    ServletSetUpObject booksGet = TestingUtil.mockGet(badParams);
+    BooksServlet booksServlet = new BooksServlet();
+    booksServlet.doGet(booksGet.getRequest(), booksGet.getResponse());
+    String output = booksGet.getWriter().toString();
+    output = TestingUtil.getInnerMap(output);
+    Assert.assertEquals(output, EMPTY_MAP_STRING);
```
seangilpin, line 35: I'm confused as to why test case returns an EMPTY_MAP_STRING and the other doesn't.

nssafir: For this test, my intention was to have a bizarre keyword that would not generate a result so the returned map is empty.

seangilpin: I don't understand how its working though. You're not actually making a call to the API correct?

nssafir: In this case, the servlet would make a call to the api.

seangilpin: Oh, ok. Usually when we make "unit test" we don't make calls to external APIs. I think you should reach out to Dan to get his opinion on this, and perhaps the VSE's.

nssafir: That makes sense. My thought with this was to show some example tests that used the entire servlet because I was working on an older branch and the specific methods in the servlets classes may be pretty different than those methods in later versions. I was planning on adding more tests for specific methods once this PR was merged and the code in the servlets is more stable. Let me know what you think of this plan.

nssafir: I just spoke with @daperlmut , and he also encouraged to do unit testing instead of integration testing (i.e. testing with API calls). I'll change the tests to reflect this and re-request a review after.

seangilpin: SG, thanks for following up with him.

PR #56, opened 2020-07-15
Use dynamic location

### LINKED ISSUE #39: Dynamic location data.
Don't save the coordinates as a static cookie. Instead query them every time the user reloads the results.

juanuribe28:
Refactor the location pipeline to use dynamic settings instead of a constant cookie. Closes #39


src/main/webapp/scripts/toggle.js
```diff
@@ -73,35 +77,55 @@ function renderKeyword(template, keywordObj) {
...
+    const civicObj = await loadObject(`${CIVIC_OBJ_URL}?address=${address}`);
+    const civicLocationObj = buildCivicLocationObj(civicObj);
+    const locationTemplate = await LOCATION_TEMPLATE_PROMISE;
+    renderLocation(locationTemplate, civicLocationObj);
+  } else {
+    alert('Sorry, your current location is not supported');
   }
 }
 
+function locationObj2Address(locationObj) {
```
nssafir, line 129: Are we printing the user's address in the website?

juanuribe28: No, but the civic API now takes an address

juanuribe28: Also, this would allow the user to input any address in the future

nssafir (approved):

src/main/java/com/google/sps/servlets/BooksServlet.java
```diff
@@ -45,9 +46,14 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     LinkedHashMap<String, List<Book>> booksMap = new LinkedHashMap<>();
     terms.forEach((term) -> {
-      String jsonString = getJsonStringForTerm(term);
-      List<Book> books = makeBooksList(jsonString);
-      booksMap.put(term, books);
+      try {
+        String jsonString = getJsonStringForTerm(term);
+        List<Book> books = makeBooksList(jsonString);
+        booksMap.put(term, books);
+      } catch(Exception e) {
```
daperlmut: [Nit] As mentioned in previous PRs I'd really try to avoid using broad catch statement when possible.  Do we have a specific error we are worried about occurring here?

src/main/java/com/google/sps/servlets/actions/ProjectsServlet.java
```diff
@@ -40,8 +40,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         String jsonResult = UrlRequest.urlQuery(API_PATH, queryParams);
         List<Project> projectsList = extractProjectsList(jsonResult);
         jsonResultMap.put(term, projectsList);
-      } catch(NullPointerException npe) {
-        npe.printStackTrace();
+      } catch(Exception e) {
```
daperlmut: Why was this broadened to catch any exception?  Are there multiple types of exceptions that can occur now?

daperlmut (changes requested):

src/main/java/com/google/sps/servlets/actions/ProjectsServlet.java
juanuribe28: Yeah, when there are no results there's a NullPointerException but when there's a problem with the API_KEY or with the API itself there's another one.

src/main/java/com/google/sps/servlets/BooksServlet.java
juanuribe28: Yeah, when there are no results there's a NullPointerException but when there's a problem with the API_KEY or with the API itself there's another one

daperlmut: I see.  In that case,  I would recommend identifying the types of exceptions you expect and putting them in a multiple catch block
e.g.
`try {
  myFunctionThatThrowsNullPointerOrIoException()
}  catch (NullPointerException|IOException exception) {
  handleException(exception)
}`

daperlmut: [Nit] Please put a space between 'catch' and '('.  

This applies here and in the other files for this PR.

src/main/java/com/google/sps/servlets/actions/ProjectsServlet.java
daperlmut: See comment above.

daperlmut (changes requested):

daperlmut (approved):

PR #59, opened 2020-07-16
Implement loading counter.

juanuribe28:
Add a loading counter to keep track of the number of elements left to load.

### DIFF
--- a/src/main/webapp/scripts/counter.js
+++ b/src/main/webapp/scripts/counter.js
@@ -1,0 +1,37 @@
+let loadingCounter = (function(value = 0) {
+
+  const getValue = function() {
+    return value;
+  }
+
+  const changeBy = function (n) {
+    if (isNaN(n)) {
+      throw `${n} is not a number`;
+    }
+    value += n;
+    hide();
+    return value;
+  }
+
+  const hide = function() {
+    if (value < 0) {
+      $('body').removeClass('loading');
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


src/main/webapp/scripts/counter.js
```diff
@@ -0,0 +1,37 @@
+let loadingCounter = (function(value = 0) {
```
nssafir, line 1: [optional] it may be more intuitive to make loadingCounter a class, but it's your call.

nssafir (approved):

src/main/webapp/scripts/counter.js
juanuribe28: In javascript it is preferred to use closures and modules over classes.

daperlmut: Juan and I discussed this today as well.  If this were Java I would support making this a Class but in Javascript I agree that the use of closures over Classes is the preferred style.

Juan, regarding our discussion of using Reflection in a callback were you able to find out any more about that?  I'll approve this for now but in a future PR it might be nice to abstract the hide functionality outside of the counter module.

daperlmut (approved):

src/main/webapp/scripts/counter.js
juanuribe28: Yeah, I just got out PR #62 tracing the counter methods.

PR #60, opened 2020-07-16
Input nits

dranilew:
This PR should fix the problem with empty form inputs (like submtiting no text or submitting no images). This should also fix the problem where less relevant keywords associated with images are put into the keywords list. 

There is still the problem where if the user inputs an invalid image file, the website will still end up failing. This will come out in a later PR.

### DIFF
--- a/src/main/java/com/google/sps/data/Keywords.java
+++ b/src/main/java/com/google/sps/data/Keywords.java
@@ -57,7 +57,9 @@ public final class Keywords {
   public static String addKeywords(List<EntityAnnotation> blobAnalysis) throws IOException {
     List<String> labelList = new ArrayList<>();
     for (EntityAnnotation label : blobAnalysis) {
-      labelList.add(label.getDescription());
+      if (label.getScore() >= 0.75) {
+        labelList.add(label.getDescription());
+      }
     }
     String labelSentence = String.join(", ", labelList);
     return addToDatastore(labelSentence);
@@ -93,7 +95,7 @@ public final class Keywords {
       }
     });
     for (com.google.cloud.language.v1.Entity entity : entities) {
-      if (entity.getSalience() == 0) {
+      if (entity.getSalience() < 0.3) {
         continue;
       }
       orderSet.add(entity);
--- a/src/main/java/com/google/sps/data/Keywords.java
+++ b/src/main/java/com/google/sps/data/Keywords.java
@@ -37,11 +37,11 @@ public final class Keywords {
   // We don't want any terms that are far too un-salient to include in the keywords list.
   // 0.2 is a good minimum threshold as, especially when extracting saliency from a list of
   // label, many of the high-relevancy terms end up only getting assigned a salience of 0.2.
-  private static final int MIN_SALIENCE_THRESHOLD = 0.2;
+  private static final double MIN_SALIENCE_THRESHOLD = 0.1;
 
   // For labels, we want high-confidence labels. Many of the labels below 0.70 seem to end
   // up being only tangetially related to the contents of the picture.
-  private static final int MIN_SCORE_THRESHOLD = 0.70;
+  private static final double MIN_SCORE_THRESHOLD = 0.60;
 
   /**
    * @return a list of the 5 most salient keywords
--- a/src/main/java/com/google/sps/data/Keywords.java
+++ b/src/main/java/com/google/sps/data/Keywords.java
@@ -34,6 +34,15 @@ public final class Keywords {
 
   private static final int MAX_NUM_KEYWORDS = 10;
 
+  // We don't want any terms that are far too un-salient to include in the keywords list.
+  // 0.2 is a good minimum threshold as, especially when extracting saliency from a list of
+  // label, many of the high-relevancy terms end up only getting assigned a salience of 0.2.
+  private static final int MIN_SALIENCE_THRESHOLD = 0.2;
+
+  // For labels, we want high-confidence labels. Many of the labels below 0.70 seem to end
+  // up being only tangetially related to the contents of the picture.
+  private static final int MIN_SCORE_THRESHOLD = 0.70;
+
   /**
    * @return a list of the 5 most salient keywords
    */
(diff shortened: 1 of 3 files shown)


src/main/webapp/index.html
```diff
@@ -26,9 +26,9 @@
         <button id="select-text">Text</button>
         <button id="select-blob">Upload Image</button>
       </div>
-      <form id="keyword-form" class="hide" action="/keyword" method="post" enctype="multipart/form-data">
+      <form id="keyword-form" class="hide" action="/keyword" method="post" enctype="multipart/form-data" onsubmit="validateForm(event);">
```
juanuribe28: [Nit] Do not combine markup with javascript. Instead of using onsubmit add an event listener.

src/main/webapp/index.html
```diff
@@ -26,9 +26,9 @@
...
         <textarea id="text-input" name="keyword-sentence" placeholder="Enter your tweet or interests here"></textarea>
-        <input id="blob-input" type="file" name="image" accept="image/*" />
+        <input id="blob-input" type="file" name="image" accept="image/png, image/jpeg, image/jpg" />
```
juanuribe28: Why are you only allowing these formats of images?

src/main/webapp/scripts/keywords.js
```diff
@@ -15,3 +15,34 @@ $().ready(function() {
...
+ * Checks whether the form inputs are valid before submitting said form.
+ */
+function validateForm(event) {
+  if ($('#blob-input').hasClass('hide')) {
+    // Check the text input
+    if ($('#text-input').val() == null || $('#text-input').val() === "") {
+      alert('Your input cannot be empty!'); 
+      event.preventDefault();
+      return false;
+    } else {}
```
juanuribe28: [Nit] Remove empty else

src/main/webapp/scripts/keywords.js
```diff
@@ -15,3 +15,34 @@ $().ready(function() {
...
+  } else {  // Check blob input
+    const blobInput = document.getElementById('blob-input');
```
juanuribe28: Why are you using vanilla javascript to get this element, and not jQuery like in the rest of the script?

juanuribe28 (changes requested):

src/main/webapp/scripts/keywords.js
dranilew: This is needed so that the next else statement does not get attributed to the wrong if statement.

src/main/webapp/index.html
dranilew: I'm not sure actually lol, I just kinda changed it out of whim. I will change it back.

src/main/webapp/scripts/keywords.js
dranilew: I didn't learn about the jquery way to do it until just now, so I will be converting it in the next commit.

src/main/webapp/scripts/keywords.js
dranilew: Just kidding, this is not needed.

src/main/webapp/scripts/keywords.js
```diff
@@ -14,4 +14,33 @@ $().ready(function() {
...
+    } else {  // Check blob input
+      const blobInput = $('#blob-input').prop('files');
+      if (blobInput.length == 0) {
+        alert("You must upload an image file!");
+        event.preventDefault();
+        return false;
+      }
+      const blobFile = blobInput[0];
+      const fileType = blobFile['type'];
+      const validImageTypes = ['image/jpeg', 'image/png', 'image/jpg'];
```
juanuribe28: [Optional] Just check whether it starts with image.

juanuribe28 (approved):

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -57,7 +57,9 @@
   public static String addKeywords(List<EntityAnnotation> blobAnalysis) throws IOException {
     List<String> labelList = new ArrayList<>();
     for (EntityAnnotation label : blobAnalysis) {
-      labelList.add(label.getDescription());
+      if (label.getScore() >= 0.75) {
```
daperlmut: [Nit] I work try to avoid magic values in the code.  For some cutoff like 0.75 I would create a constant with a description name and maybe add a comment on how you decided on this value.

The same is true of 0.3 below.

src/main/webapp/scripts/keywords.js
```diff
@@ -14,4 +14,32 @@ $().ready(function() {
     $('#blob-input').addClass('hide').val(null);
     $('#form-title *').text('Copy Your Tweet Here');
   });
+
+  $('#keyword-form').on('submit', function(event) {
+    // Check whether the inputs are valid before submitting.
+    if ($('#blob-input').hasClass('hide')) {
+      // Check the text input
```
daperlmut: [Nit] Add a '.'

src/main/webapp/scripts/keywords.js
```diff
@@ -14,4 +14,32 @@ $().ready(function() {
...
+      if ($('#text-input').val() == null || $('#text-input').val() === "") {
+        alert('Your input cannot be empty!'); 
+        event.preventDefault();
+        return false;
+      }
+    } else {  // Check blob input
```
daperlmut: [Nit] Add a '.'

daperlmut (changes requested):

src/main/java/com/google/sps/data/Keywords.java
seangilpin: Maybe put a comment here explaining you only want to add high confidence labels.... (I assume thats what this is doing?)

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -93,7 +95,7 @@ private static String addToDatastore(String message) throws IOException {
       }
     });
     for (com.google.cloud.language.v1.Entity entity : entities) {
-      if (entity.getSalience() == 0) {
+      if (entity.getSalience() < 0.3) {
```
seangilpin: These filter thresholds should generally be stored in a constant.

src/main/webapp/scripts/keywords.js
seangilpin: Cleaner comments if you mention this block checks if any file uploaded, and in the next block you check if a valid image file is uploaded.

src/main/java/com/google/sps/data/Keywords.java
dranilew: Yes.

daperlmut (approved):

PR #62, opened 2020-07-16
Implement progress bar

juanuribe28:
Add a progress bar using the counter, to keep track of keywords loading.

### DIFF
--- a/src/main/webapp/scripts/toggle.js
+++ b/src/main/webapp/scripts/toggle.js
@@ -25,7 +25,7 @@ async function loadContentSection() {
   } else {
     keywords.forEach(loadKeywordSection);
   }
-  setTimeout(loadCivicSection, 5000);
+  loadCivicSection();
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/webapp/scripts/toggle.js
```diff
@@ -9,14 +9,17 @@ const BOOKS_OBJ_URL =  '/books';
 const PROJECTS_OBJ_URL = '/actions/projects';
 const OBJECTS_URLS = [BOOKS_OBJ_URL, PROJECTS_OBJ_URL];
 
+let loadingCounter;
```
dranilew, line 12: What's the purpose of this variable other than just getting updated with a new value?

juanuribe28: The purpose is to have a global loading decorator for the regular counter. Every time an object finishes loading and it's rendered it calls loadingCounter.decrement(). The decorator allows to update the progress bar and hide the loading cursor when completed.

dranilew (approved):

daperlmut (approved):

PR #66, opened 2020-07-20
Back End for News

nssafir:
Implemented news servlet to scrape news articles from Google News. Still needs to be integrated with front end.

PR #69, opened 2020-07-22
Implement Testing Framework for News Servlet

nssafir:
Implement testing framework for NewsServlet class using JUnit and Mockito. Test getHTML and makeArticleList methods, will likely move encodeMapAsJson method to Util class in later PR.


src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,115 @@
...
+import java.net.HttpURLConnection;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+/** Servlet that gets news information. */
+@WebServlet("/news")
+public class NewsServlet extends HttpServlet {
+
+  public final String GOOGLE_NEWS_PATH = "https://news.google.com/search";
+  public final String articleTag = "DY5T1d"; // Class tag for articles in Google News html code.
```
dranilew: [Nit] Use constant case for constant instance variables. So in this case, you'd call the variable `ARTICLE_TAG` instead of `articleTag`

src/main/java/com/google/sps/servlets/NewsServlet.java
```diff
@@ -0,0 +1,115 @@
...
+    return contentBuilder.toString();
+  }
+
+  public List<Article> makeArticleList(String HTMLString) throws IOException {
+    List<Article> articleList = new ArrayList<>();
+    String[] substrings = HTMLString.split(articleTag);
+    String thisSubstring, lastSubstring, title, link;
+    int closeBracket, openBracket;
+    // For each substring seperated by the article class tag.
+    for (int i = 4; i < Math.min(substrings.length, 4 + MAX_ARTICLES); i++) {
```
dranilew, line 96: What's the reason behind starting the for-loop at index 4?

src/test/java/com/google/sps/tests/NewsTest.java
```diff
@@ -0,0 +1,145 @@
...
+    try {
+      HttpURLConnection connection = Mockito.mock(HttpURLConnection.class);
+      Mockito.when(connection.getInputStream()).thenReturn(new FileInputStream(FULL_HTML_PATH));
+      NewsServlet ns = new NewsServlet();
+      Assert.assertEquals(ns.getHTML(connection), textToString(FULL_HTML_PATH));  
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
```
dranilew: [Nit] Remove new line

src/test/java/com/google/sps/tests/NewsTest.java
```diff
@@ -0,0 +1,145 @@
...
+      "news.google.com/articles/CBMiVmh0dHBzOi8vd3d3LnNjaWVuY2VtYWcub3JnL25ld3MvMjAyMC8wNy9ob3ctYmx1ZS1wcm90ZWluLXR1cm5zLXRyZWUtZnJvZ3MtYnJpZ2h0LWdyZWVu0gEA?hl=en-US&amp;gl=US&amp;ceid=US%3Aen"
+    ));
+  }
+
+ /**
+  * Check that when given no HTML text, makeArticleList returns an empty list.
+  */
+  @Test
+  public void articlesFromEmptyHTML() throws IOException {
+    NewsServlet ns = new NewsServlet();
```
dranilew: [Optional] You can consider using a method with the @BeforeEach tag to initialize a class-level NewsServlet variable. That way a new NewsServlet is created and initialized before every test method (marked by @Test).

dranilew (changes requested):

src/test/java/com/google/sps/tests/NewsTest.java
nssafir: I did not know about this tag before, thanks!

src/main/java/com/google/sps/servlets/NewsServlet.java
nssafir: I just addressed this in some changes for PR #66

src/test/java/com/google/sps/tests/NewsTest.java
nssafir: Upon further research, the @beforeeach tag is just @before for junit 5. In junit4, the @before tag runs once before each individual test, while the @beforeclass tag runs once before all the tests (i.e., if you had a class with ten tests, @before would run ten times and @beforeclass would run once).

dranilew (approved):

src/main/java/com/google/sps/servlets/NewsServlet.java
daperlmut: At this point I think it's worth making 4 into a constant.

src/test/java/com/google/sps/tests/NewsTest.java
```diff
@@ -0,0 +1,152 @@
...
+import java.io.FileInputStream;
+import java.net.HttpURLConnection;
+import java.util.Map;
+import java.util.HashMap;
+
+
+@RunWith(JUnit4.class)
+public final class NewsTest {
+
+  public final static String FULL_HTML_PATH = "src/test/java/com/google/sps/data/fullNewsHTML.txt";
```
daperlmut: [Nit] Since you use the string 'src/test/java/com/google/sps/data' in all of the constants here I would just create a constant for that.

`public final static String HTML_PATH_PREFIX = "src/test/java/com/google/sps/data"`

Then you can construct the other constants by just concatenating onto the end of that prefix.

src/test/java/com/google/sps/tests/NewsTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+ /**
+  * Helper method to check the makeArticleList method in the NewsServlet class.
+  * Checks if makeArticleList on HTML text in the specified file has output equivalent to expectedList.
+  */
+  private void checkMakeArticleList(String filePath, List<Article> expectedList) throws IOException {
+    String htmlString = textToString(filePath);
+    List<Article> articleList = ns.makeArticleList(htmlString);
+    // Check each class instance in each article object is equal.
+    Assert.assertEquals(articleList.size(), expectedList.size());
```
daperlmut, line 62: [Optional] Theoretically I think there are fancier ways of doing this.  At least in c++ I know there is a way to define an equality method on options and then there is an assertListEquals helper.

You may want to look into junit.Assert which has some particularly fancy helper methods.  See https://stackoverflow.com/questions/3236880/assert-equals-between-2-lists-in-junit for an example.

daperlmut: Oh, and I had a typo above, I meant to say "and equality method on objects" not "on options"

src/test/java/com/google/sps/tests/NewsTest.java
```diff
@@ -0,0 +1,152 @@
...
+  */
+  @Test
+  public void articlesFromEmptyHTML() throws IOException {
+    List<Article> articleList = ns.makeArticleList("");
+    Assert.assertEquals(articleList, new ArrayList<Article>());
+  }
+
+ /**
+  * Checks that makeArticleList makes a list with all the articles in the HTML text
+  * if there are less articles in the HTML text than makeArticleList would genereate by default.
```
daperlmut: [Optional] Line wrapping looks a bit weird here.  Can't you find more words on the first line w/o going over the 80 character limit?

daperlmut (changes requested):

src/test/java/com/google/sps/tests/NewsTest.java
nssafir: I tried looking into this for a couple hours, but nothing in the JUnit library or Hamcrest library was working for me.

daperlmut: Ok, you can leave it as is but can you just confirm to me that the following doesn't work
`assertThat(articleList, is(expectedList))`

daperlmut (approved):

src/test/java/com/google/sps/tests/NewsTest.java
nssafir: Yes, that is not working.

nssafir: I tried Juan's method but it also did not work for Classes instead of Strings.

PR #72, opened 2020-07-23
Uses client's country code as param for news servlet.

nssafir:
Request for news servlet now takes user's coordinates. The coordinates are decoded to get the clients country, which is then used as a paramater in the Google News url to boost results from that country.


src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -96,6 +97,9 @@ private void jsonToMap(String jsonString) {
             case "route" :
               geoMap.put("Street Name", name);
               break;
+            case "country" :
+              geoMap.put("Country", shortName);
+              break;
```
juanuribe28: toogle.js is using full country name in line 119, so if we're going to be using the short name now it would be useful to change that too.

juanuribe28 (approved):

daperlmut (approved):

PR #86, opened 2020-07-28
Tests for ProjectsServlet.java

juanuribe28:
Add tests for the projects servlet.

### DIFF
--- a/src/test/java/com/google/sps/tests/ProjectsServletTest.java
+++ b/src/test/java/com/google/sps/tests/ProjectsServletTest.java
@@ -1,0 +1,105 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import org.powermock.api.mockito.PowerMockito;
+import com.google.sps.data.UrlRequest;
+
+/** */
+@RunWith(PowerMockRunner.class) 
+@PrepareForTest({ UrlRequest.class, ProjectsServlet.class })
+public final class ProjectsServletTest { 
+
+  private final static ProjectsServlet PROJECTS_SERVLET = new ProjectsServlet();
+  private final static String[] KEYWORD = {"keyword"};
+
+  private void setUpUrlQueryMock(String response) throws IOException {
+    PowerMockito.mockStatic(UrlRequest.class);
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/sps/tests/ProjectsServletTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private final static ProjectsServlet PROJECTS_SERVLET = new ProjectsServlet();
+  private final static String[] KEYWORD = {"keyword"};
+
+  private void setUpUrlQueryMock(String response) throws IOException {
+    PowerMockito.mockStatic(UrlRequest.class);
+    PowerMockito.when(UrlRequest.urlQuery(Mockito.anyString(), Mockito.any(Map.class))).thenReturn(response);
+  }
+
+  private String mockServlet() throws IOException {
```
nssafir, line 32: [Optional] can probably delete some of the whitespace in this method

src/test/java/com/google/sps/tests/ProjectsServletTest.java
```diff
@@ -0,0 +1,105 @@
...
+    PROJECTS_SERVLET.doGet(mockedRequest, mockedResponse);
+
+    Mockito.verify(mockedRequest, Mockito.atLeast(1)).getParameterValues("key");
+    writer.flush();
+    return stringWriter.toString();
+  }
+
+  @Test
+  public void testServlet() throws IOException {
+    String apiResponse = "{\"search\":{\"response\":{\"projects\":{\"project\":[{"
```
nssafir, line 49: [Nit] can declare these long strings as const

juanuribe28: I'm only using them once, why should I make them global const.

nssafir: My bad, you're right.

nssafir (approved):

src/test/java/com/google/sps/tests/ProjectsServletTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    PROJECTS_SERVLET.doGet(mockedRequest, mockedResponse);
+    Mockito.verify(mockedRequest, Mockito.atLeast(1)).getParameterValues("key");
+    writer.flush();
+    return stringWriter.toString();
+  }
+
+  @Test
+  public void testServlet() throws IOException {
+    String apiResponse = "{\"search\":{\"response\":{\"projects\":{\"project\":[{"
```
daperlmut, line 49: [Nit] I think you might be able to do this more readably using raw string 
https://docs.w3cub.com/javascript/global_objects/string/raw/

daperlmut (changes requested):

src/test/java/com/google/sps/tests/ProjectsServletTest.java
juanuribe28: I wasn't able to find raw strings for Java

juanuribe28: I think they're only for Java 12

daperlmut: Ah, I see.  I got confused because the version of javascript you are using supports raw string but I guess Java 8 doesn't.  This is fine then.

daperlmut (approved):

Issue #74, opened 2020-07-23
All Caps  keywords


juanuribe28:
Make keywords all upper case instead of all lower case.

PR #89, opened 2020-07-28
Implement functional user log-in with a basic front-end

dranilew:
The front-end is very basic, and will require a lot more work to make this look better. For the time being, however, this provides a basic yet functional way for users to access their own previous set of keywords (no favorite option yet). Upcoming is, of course, a lot of front-end work to make this more intuitive and look better.


src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -114,8 +122,11 @@ private static String addToDatastore(String message) throws IOException {
     for (com.google.cloud.language.v1.Entity entity : orderSet) {
       keywordSet.add(entity.getName().toLowerCase());
     }
+    UserService userService = UserServiceFactory.getUserService();
+    final String userId = userService.getCurrentUser().getUserId();
     Entity datastoreEntity = new Entity("Keyword");
     datastoreEntity.setProperty("keywords", keywordSet);
+    datastoreEntity.setProperty("id", userId);
```
juanuribe28, line 127: [Optional] You could make it so the userId is also the key

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -36,27 +40,31 @@
...
-      return keywordList;
-    } catch (EntityNotFoundException ex) {
-      return new ArrayList<>();
-    }
+  public static Collection<String> getKeywords(String keyString) {
+    UserService userService = UserServiceFactory.getUserService();
+    Map<String, Collection<String>> allKeywords = getKeyKeywordMap(userService.getCurrentUser().getUserId());
+        
+    Collection<String> keywordCollection = allKeywords.get(keyString);
+    List<String> keywordList = keywordCollection.stream().limit(MAX_NUM_KEYWORDS).collect(Collectors.toList());
```
juanuribe28: [Optional] You can add this limit to the datastore query itself.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -36,27 +40,31 @@
...
+  private static final double MIN_SALIENCE_THRESHOLD = 0.08;
 
-  // For labels, we want high-confidence labels. Many of the labels below 0.70 seem to end
+  // For labels, we want high-confidence labels. Many of the labels below 0.60 seem to end
   // up being only tangetially related to the contents of the picture.
   private static final double MIN_SCORE_THRESHOLD = 0.60;
 
   /**
-   * @return a list of the 5 most salient keywords
+   * @return a list of the 10 most salient keywords from the datastore in accordance to the keys
```
juanuribe28: [Nit] Add dot at the end.

src/main/java/com/google/sps/servlets/UserServlet.java
```diff
@@ -0,0 +1,40 @@
...
+      String loginUrl = userService.createLoginURL("/");
+      logInfo.add(loginUrl);
+      logInfo.add("placeholder");
+    }
+    Gson gson = new Gson();
+    String json = gson.toJson(logInfo);
+    response.getWriter().println(json);
+  }
+}
+
```
juanuribe28: [Nit] Remove new line

src/main/webapp/index.html
```diff
@@ -28,6 +28,9 @@
         <div id="form-title">
           <p>Choose Type of Input</p>
         </div>
+        <div id="log-btn" class="hide">
+          <a id="log-btn" class="hide button"></a>
```
juanuribe28: [Nit] Ids should be unique. Use "log-button" as class.

src/main/webapp/scripts/keywords.js
```diff
@@ -1,6 +1,47 @@
 const BLOB_URL_PROMISE = loadTemplate('/user-image');
 
 $().ready(function() {
+  fetch('/login').then(response => response.json()).then((logURL) => {
```
juanuribe28: [Optional] Instead of fetch and chaining promises use load.js functions to load object and then use await instead of .then. You would also have to mark the callback function as async.

juanuribe28: The same way it is being done for '/user-image' and the BLOB_URL_PROMISE.

src/main/webapp/scripts/keywords.js
```diff
@@ -1,6 +1,47 @@
...
+      $('#keyword-form, #input-select, #location-input, #previous-keywords').addClass('hide');
+    } else {
+      $('#form-title > p').text('Choose Type of Input');
+      $('#log-btn > a').text('Logout');
+      $('#input-select, #location-input, #previous-keywords').removeClass('hide');
+    }
+    $('#log-btn').removeClass('hide');
+  });
+
+  fetch('/all-keywords').then(response => response.json()).then((keywordMap) => {
```
juanuribe28: [Optional] Same as above. Preload the object and use async-await keywords instead of chaining promises.

src/main/webapp/scripts/keywords.js
```diff
@@ -54,3 +95,10 @@ $().ready(function() {
     return true;
   });
 });
+
+function addKeyword(key, value, map) {
+  let btn = $('<a></a>').attr('href', '/results?k=' + key);
+  let keywords = value.join(', ');
+  btn.text(keywords);
+  keywordsDiv.append(btn);
+}
```
juanuribe28: Are you using this function?

juanuribe28 (changes requested):

src/main/webapp/scripts/keywords.js
dranilew: I forgot to remove this function, thanks for pointing this out!

src/main/webapp/scripts/keywords.js
dranilew: I will look into this.

src/main/java/com/google/sps/data/Keywords.java
dranilew: I guess I could move this into the addToDatastore method itself, so I can just grab and return.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -110,12 +115,21 @@ private static String addToDatastore(String message) throws IOException {
       orderSet.add(entity);
     }
     // We must create a new list of Strings as a collection of NLP Entities is not supported by datastore.
+    /**
     Set<String> keywordSet = new HashSet<>();
     for (com.google.cloud.language.v1.Entity entity : orderSet) {
       keywordSet.add(entity.getName().toLowerCase());
     }
+    */
+    Collection<String> keywordCollection =
```
nssafir, line 119: [Optional] this stream is a little hard to follow, maybe add a comennt

src/main/java/com/google/sps/servlets/UserServlet.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/login")
+public class UserServlet extends HttpServlet {
+
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
+
```
nssafir: [Nit] delete newline

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -110,12 +115,16 @@ private static String addToDatastore(String message) throws IOException {
       orderSet.add(entity);
     }
     // We must create a new list of Strings as a collection of NLP Entities is not supported by datastore.
-    Set<String> keywordSet = new HashSet<>();
-    for (com.google.cloud.language.v1.Entity entity : orderSet) {
-      keywordSet.add(entity.getName().toLowerCase());
-    }
+    // We can limit the collection to 10 keywords, map every entity to their names, and only allow distinct values. 
+    Collection<String> keywordCollection = 
+        orderSet.stream().limit(MAX_NUM_KEYWORDS).map(e -> e.getName().toLowerCase()).distinct().collect(Collectors.toList());
```
juanuribe28, line 120: I would be interesting to use upper case instead #74

juanuribe28 (approved):

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -136,5 +145,22 @@ private static AnalyzeEntitiesResponse analyzeEntity(String message) throws IOEx
...
       return null;
     }
-  }  
+  }
+  
+  /**
+   * @return a map of keys to their list of keywords of all sets of keywords that the
+   * user defined by id has entered.
+   */
+  private static Map<String, Collection<String>> getKeyKeywordMap(String id) {
```
daperlmut: [Nit] I might call this KeyToKeywordMap as I think it's a bit easier to parse without having 'Key' twice in a row.

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -136,5 +145,22 @@ private static AnalyzeEntitiesResponse analyzeEntity(String message) throws IOEx
       ex.printStackTrace();
       return null;
     }
-  }  
+  }
+  
+  /**
+   * @return a map of keys to their list of keywords of all sets of keywords that the
```
daperlmut: I'm having a bit of trouble parsing "their list of keywords of all sets of keywords".  Can you clarify how the list of keywords associated with each key is chosen.

daperlmut (changes requested):

src/main/java/com/google/sps/data/Keywords.java
dranilew: The list of keywords associated to each key is gotten from the NLP API, which extracts salient terms from the user's input. The key is automatically generated upon creation of a new Entity object (which we do before setting its properties and putting the entity into the datastore for later retrieval), so we use that automatically-generated key instead. But, I do agree that the Javadoc is confusing, and I will make it more clear what it does.

daperlmut (approved):

PR #94, opened 2020-07-29
Implement Cookie System for Location

nssafir:
Store user coordinates as cookies again, but make sure to erase when location checkbox is unchecked. The location integration feature for news now works again, and it is faster because the location servlet is called only once total.

### DIFF
--- a/src/main/webapp/scripts/index.js
+++ b/src/main/webapp/scripts/index.js
@@ -2,13 +2,19 @@ $().ready(function() {
   const location = getCookie('location');
   if (location != "") {
     $('#location-checkbox').prop('checked', true);
+    setCoordinateCookie();
+  }
+  else {
+    eraseCookie('coordinates');
   }
 
   $('#location-checkbox').change(function() {
     if(this.checked) {
       setCookie('location', 'true', 1);
+      setCoordinateCookie();
     } else {
       eraseCookie('location');
+      eraseCookie('coordinates');
     }
   });
   $('#address').val(getCookie("address"));
--- a/src/main/webapp/scripts/toggle.js
+++ b/src/main/webapp/scripts/toggle.js
@@ -12,6 +12,8 @@ const PROJECTS_OBJ_URL = '/actions/projects';
 const OBJECTS_URLS = [NEWS_OBJ_URL, BOOKS_OBJ_URL, PROJECTS_OBJ_URL];
 
 let loadingCounter;
+let lat = getCookie("lat");
+let lng = getCookie("lng");
 
 /**
  * Loads the content section. 
@@ -33,7 +35,10 @@ async function loadContentSection() {
     loadCivicSectionFromAddress(address);
   } else {
     if (location != "") {
-      loadLocationObj(loadCivicSectionFromLocation);
+      console.log(`${LOCATION_OBJ_URL}?lat=${lat}&lng=${lng}`);
+      let locationObj = await loadObject(`${LOCATION_OBJ_URL}?lat=${lat}&lng=${lng}`);
+      console.log(locationObj);
+      loadCivicSectionFromLocation(locationObj);
     } else {
       loadingCounter.decrement();
     }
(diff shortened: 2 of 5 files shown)


src/main/webapp/scripts/index.js
```diff
@@ -2,13 +2,19 @@ $().ready(function() {
...
+    eraseCookie('coordinates');
   }
 
   $('#location-checkbox').change(function() {
     if(this.checked) {
       setCookie('location', 'true', 1);
+      setCoordinateCookie();
     } else {
       eraseCookie('location');
+      eraseCookie('coordinates');
```
juanuribe28, line 17: I don't think we need the location cookie, we can just check whether coordinates exists.

src/main/webapp/scripts/index.js
```diff
@@ -2,13 +2,19 @@ $().ready(function() {
   const location = getCookie('location');
   if (location != "") {
     $('#location-checkbox').prop('checked', true);
+    setCoordinateCookie();
+  }
+  else {
+    eraseCookie('coordinates');
```
juanuribe28, line 8: Where are you setting up this cookie? I think you should be erasing lat and lng instead.

src/main/webapp/scripts/toggle.js
```diff
@@ -18,6 +20,12 @@ let loadingCounter;
  * Returns a promise that resolves when everything loads.
  */
 async function loadContentSection() {
+  let location = getCookie('location');
+  if (location != "") {
+    let lat = getCookie("lat");
+    let lng = getCookie("lng");
+    locationObj = await loadObject(`${LOCATION_OBJ_URL}?lat=${lat}&lng=${lng}`);
```
juanuribe28, line 27: I think we could make this call in the main page, and save lat, lng and address as a cookie array. That way we don't have to load the same object multiple times.

src/main/webapp/scripts/toggle.js
```diff
@@ -28,15 +36,12 @@ async function loadContentSection() {
     keywords.forEach(loadKeywordSection);
   }
   const address = encodeURI(getCookie('address'));
-  const location = getCookie('location');
   if (address != "") {
     loadCivicSectionFromAddress(address);
+  } else if (location != "") {
+    loadCivicSectionFromLocation(locationObj);
```
juanuribe28, line 42: If we always save the address as mentioned above we could load the civic section from address all the time.

src/main/webapp/scripts/toggle.js
```diff
@@ -70,12 +75,20 @@ function hideLoading() {
   $("#real-body").removeClass("hide").addClass("body");
 }
 
+function makeUrl(url, keyword) {
+  let queryString = `?key=${keyword}`;
+  let fullUrl = `${url}${queryString}`;
+  if (url === '/news' && locationObj != null) {
+    fullUrl = `${fullUrl}&country=${locationObj["Short Country"]}`;
```
juanuribe28, line 82: [Optional] Using dot notation instead of []. This would require removing the space from 'Short Country'

src/main/webapp/scripts/toggle.js
```diff
@@ -105,23 +118,16 @@ function renderKeyword(template, keywordObj) {
...
-      callback(locationObj);
-    }, (err) => {
-      console.warn(`ERROR(${err.code}): ${err.message}`);
-      alert("We cannot access your location. Try checking your browswer settings");
-    });
-  }
+  locationObj = await loadObject(`${LOCATION_OBJ_URL}?lat=${coordinates[0]}&lng=${coordinates[1]}`);
+  callback(locationObj);
 }
+*/
```
juanuribe28, line 126: [Nit] Remove if not using it.

src/main/webapp/scripts/toggle.js
nssafir: I was thinking about that earlier -- I would think it would save time, but how do you store a cookie array? I thought that all cookies had to be String objects.

src/main/webapp/scripts/toggle.js
nssafir: Would 'Short_Country' be the most stylistically consistent?

PR #101, opened 2020-07-30
Move log/logout buttons to the upper right-hand corner of the screen

dranilew:
I also spotted and fixed a typo along the way.

### DIFF
--- a/src/main/webapp/scripts/logButton.js
+++ b/src/main/webapp/scripts/logButton.js
@@ -1,9 +1,9 @@
 const LOG_URL_PROMISE = loadObject('/login');
 
 $().ready(async function() {
-  const logURL = await LOG_URL_PROMISE;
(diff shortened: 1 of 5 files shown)


juanuribe28 (approved):

src/main/webapp/scripts/logButton.js
```diff
@@ -0,0 +1,16 @@
+const LOG_URL_PROMISE = loadObject('/login');
+
+$().ready(async function() {
+  const logURL = await LOG_URL_PROMISE;
+  $('#log-btn').attr('href', logURL[0]);
+  if (logURL.length == 2) {  // User is not loggeed in.
```
daperlmut: [Nit] Noted this in a previous PR but there is a typo here.

daperlmut (changes requested):

daperlmut (approved):

PR #102, opened 2020-07-30
Implement delete keywords sets

juanuribe28:
The user is now capable of erasing any keyword set that appears in the previous keywords modal window.


dranilew (approved):

src/main/java/com/google/sps/servlets/DeleteKeywordsServlet.java
```diff
@@ -0,0 +1,25 @@
...
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+
+/** Servlet that deletes the keywords.*/
```
daperlmut: [Nit] Add a space before '*/'

src/test/java/com/google/sps/tests/DeleteKeywordsServletTest.java
```diff
@@ -0,0 +1,58 @@
...
+    DELTE_KEYWORDS_SERVLET.doPost(mockedRequest, mockedResponse);
+    Mockito.verify(mockedRequest, Mockito.atLeast(1)).getParameter("k");
+  }
+
+  @Test
+  public void testServlet() throws IOException {
+    DatastoreService mockedDatastoreService = PowerMockito.mock(DatastoreService.class);
+    setUpDatastoreMock(mockedDatastoreService);
+
+    String expected = "ahpzfnN0ZXAyMDIwLXRlYW0zNi1jYXBzdG9uZXIUCxIHS2V5d29yZBiAgICYtMiZCgw";
```
daperlmut, line 49: Is this just a random string?  I'm confused as to why we have such a long and complicated value here.

daperlmut (changes requested):

src/test/java/com/google/sps/tests/DeleteKeywordsServletTest.java
juanuribe28: It is a real key string. They have to encode the entity and the id.

daperlmut: Hmm, it looks a bit odd.  Maybe at the least we should add a comment to indicate what this is an encoding of.

daperlmut (changes requested):

daperlmut (approved):

PR #105, opened 2020-07-30
Results block

### LINKED ISSUE #91: Block results page if logged out
Block the results page and redirect to the main page if the results are not owned by the user that's logged in.

dranilew:
This should successfully block the results page from loading and popping up with the keywords when the user is not logged in.

### DIFF
--- a/src/main/webapp/scripts/toggle.js
+++ b/src/main/webapp/scripts/toggle.js
@@ -19,6 +19,11 @@ let loadingCounter;
  */
 async function loadContentSection() {
   const keywords = await loadKeywords(KEYWORDS_OBJ_URL);
+  if (keywords == null) {
+    $('#login-error').removeClass('hide');
+    hideLoading();
+    return;
+  }
   const elementsToLoad = Math.max(keywords.length, 1);
   counter.add(elementsToLoad);
   loadingCounter = traceCounterMethods(counter, elementsToLoad);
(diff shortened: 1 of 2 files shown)


juanuribe28 (approved):
[Optional] you could also block it directly from the backend as it is being served by the Results Servlet.

src/main/webapp/scripts/civic.js
```diff
@@ -7,15 +7,15 @@
...
+  const levels = {"special" : "Special",
                   "subLocality1" : "Sublocality 1",
-                  "special" : "Special",};
+                  "subLocality2" : "Sublocality 2",
+                  "locality" : "Locality",
+                  "administrativeArea2" : "County",
+                  "administrativeArea1" : "State",
+                  "regional" : "Regional",
+                   "country" : "Country",
+                  "international" : "International"};
```
daperlmut: Didn't these changes already get made in another PR?  Were there some merge issues?

daperlmut (changes requested):

src/main/webapp/scripts/civic.js
dranilew: Yea, I merged master into this branch, but I don't think master was merged into the base branch (modal-window-integration), so that's why this change is popping up.

daperlmut (approved):

PR #113, opened 2020-08-04
Fix loading bar when using location

juanuribe28:


src/main/webapp/scripts/toggle.js
```diff
@@ -24,11 +24,6 @@ async function loadContentSection() {
     locationObj = JSON.parse(getCookie('location'))
   }
   const keywords = await loadKeywords(KEYWORDS_OBJ_URL);
-  if (keywords == null) {
-    $('#login-error').removeClass('hide');
-    hideLoading();
-    return;
-  }
```
dranilew: I'm wondering whether you removed this. Had this in to initially block results page if the user isn't logged in.

juanuribe28: That's part of a different PR, where I'm trying to solve a bug

dranilew (approved):

daperlmut (approved):

PR #115, opened 2020-08-04
Implement login validation in backend for results page.

juanuribe28:
Instead of validating the user after the results page has been served, the user validation happens in the backend. This way the client only receives the information that they are supposed to be seeing.

### DIFF
--- a/src/main/java/com/google/sps/servlets/ResultsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ResultsServlet.java
@@ -10,6 +10,8 @@ import java.util.HashMap;
 import com.github.mustachejava.Mustache;
 import com.github.mustachejava.MustacheFactory;
 import com.github.mustachejava.DefaultMustacheFactory;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
 
 /** Serves the html template used in the results page with a datastore key. */
 @WebServlet("/results")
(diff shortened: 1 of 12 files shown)


src/test/java/com/google/sps/tests/ResultsServletTest.java
```diff
@@ -54,16 +72,35 @@ private String mockServlet(String keyValue) throws IOException {
...
+    boolean isUserLoggedIn = true;
+    setUpUserServiceMock(isUserLoggedIn, logoutURL);
+    String actual = mockServlet(key);
+    String expected = String.format(LOGGED_IN_RESULT_TEMPLATE, key, logoutURL);
+    Assert.assertTrue(actual.contains(expected));
+  }
+
+  @Test
+  public void testLoggedOut() throws IOException {
+    String key = "value";
```
nssafir, line 87: [Optional] because you use key and logoutUrl in a few tests with these values, you can maybe declare them global constants.

src/main/java/com/google/sps/servlets/ResultsServlet.java
```diff
@@ -10,18 +10,30 @@
...
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
     String key = request.getParameter("k");
     Map<String, String> map = new HashMap<>();
-    map.put("key", key);
     DefaultMustacheFactory mf = new DefaultMustacheFactory();
-    Mustache mustache = mf.compile("results.html");
+    Mustache mustache;
+	  if (userService.isUserLoggedIn()) {
```
dranilew, line 27: [Nit] remove extra tab.

dranilew (changes requested):

PR #116, opened 2020-08-04
Detect Language of Input

nssafir:
Detect the language of the input message (for text). Convert the language to a corresponding country, and country as a parameter of the news servlet if there is not already a country provided from the location.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
         <artifactId>appengine-maven-plugin</artifactId>
         <version>2.2.0</version>
         <configuration>
-          <deploy.projectId>step2020-team36-capstone</deploy.projectId>
+          <deploy.projectId>nsafir-step</deploy.projectId>
           <deploy.version>1</deploy.version>
         </configuration>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
         <artifactId>appengine-maven-plugin</artifactId>
         <version>2.2.0</version>
         <configuration>
-          <deploy.projectId>nsafir-step</deploy.projectId>
+          <deploy.projectId>step2020-team36-capstone</deploy.projectId>
           <deploy.version>1</deploy.version>
         </configuration>
       </plugin>
--- a/src/main/java/com/google/sps/data/Keywords.java
+++ b/src/main/java/com/google/sps/data/Keywords.java
@@ -52,9 +52,24 @@ public final class Keywords {
    */
   public static Collection<String> getKeywords(String keyString) {
     UserService userService = UserServiceFactory.getUserService();
-    Map<String, Collection<String>> allKeywords = getKeyToKeywordMap(userService.getCurrentUser().getUserId());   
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -85,7 +85,7 @@
         <artifactId>appengine-maven-plugin</artifactId>
         <version>2.2.0</version>
         <configuration>
-          <deploy.projectId>step2020-team36-capstone</deploy.projectId>
+          <deploy.projectId>nsafir-step</deploy.projectId>
```
juanuribe28: Change it back

src/main/webapp/scripts/toggle.js
```diff
@@ -23,7 +24,9 @@ async function loadContentSection() {
   if (locationCookie != "") {
     locationObj = JSON.parse(getCookie('location'))
   }
-  const keywords = await loadKeywords(KEYWORDS_OBJ_URL);
+  let keywordsJson = await loadKeywords(KEYWORDS_OBJ_URL);
```
juanuribe28: [Optional] Use const to show that you're not changing it.

src/main/webapp/scripts/toggle.js
```diff
@@ -57,7 +60,7 @@ async function loadKeywords(keywordsUrl) {
   const keywordsCookie = getCookie(key);
   if (keywordsCookie === "") {
     keywords = await loadObject(`${keywordsUrl}?k=${key}`);
-    const keywordsJson = JSON.stringify(keywords);
+    let keywordsJson = JSON.stringify(keywords);
```
juanuribe28: Why did you change it from const to let?

src/main/webapp/scripts/toggle.js
juanuribe28: [Otional] By convention I have used suffix Json if it is a json string and Obj if it is an object. So in this case I would use keywordsObj instead of keywordsJson

juanuribe28 (changes requested):

juanuribe28 (approved):

src/main/java/com/google/sps/data/Keywords.java
```diff
@@ -52,10 +52,26 @@
...
+    UserService userService = UserServiceFactory.getUserService();
+    String id = userService.getCurrentUser().getUserId();
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Keyword").setFilter(
+        new Query.FilterPredicate("id", Query.FilterOperator.EQUAL, id));
+    PreparedQuery results = datastore.prepare(query);
+    Map<String, String> keyKeywordMap = new HashMap<>();
+    for (Entity entity : results.asIterable()) {
+      String language = (String) entity.getProperty("language");
+      // Map the key to its respective collection of keywords
```
daperlmut: [Nit] Add a '.'

daperlmut (changes requested):

daperlmut (approved):

